Issue #7, opened 2025-05-22
Store retrieval results in a database


pyropy:
Upon the retrieval attempt worker should return the response to the client and record the retrieval stats in some type of database to be reported afterwards.

Possible way of storing the data could be [Cloudflare D1 Database](https://developers.cloudflare.com/d1/)

Blocked by #5

juliangruber:
Related: https://developers.cloudflare.com/logs/

pyropy:
Related to #13

PR #8, opened 2025-05-22
add: Retrieval worker

### LINKED ISSUE #5: Implement Retrieval Cloudflare Worker
Implement Cloudflare Worker that will perform retrieval of proof-set pieces from storage providers supporting the PDP. Initially users should provide proof-set ID and storage provider URL that retrieval needs to be performed from.

Blocked by #2

pyropy:
This pull requests introduces cloudflare retrieval worker. Retrieval worker is responsible for fetching the content from storage provider(s) and caching the response.

By default content is cached for responses with status code in range 200 - 299 with default time-to-live set to 24hrs (86400 seconds). This is configurable via the `CACHE_TTL` environment variable passed down to the cloudflare worker. Responses with status codes 404 and in range 500 - 599 are currently not cached (subject to change).

This pull request also introduces changes in the deployment and testing workflows, change to the main config file and updates to the README to reflect these changes.

Closes #5


bin/worker.js
```diff
@@ -1,5 +1,49 @@
+import { retrieveFile as defaultRetrieveFile } from '../lib/retrieval.js'
+
 export default {
-  async fetch (request, env, ctx) {
-    return await fetch(request)
+  async fetch (request, env, ctx, { retrieveFile = defaultRetrieveFile } = {}) {
```
NikolasHaimerl, line 4: I don't see the worker using the cache here. Is this something that you plan on adding in a follow up PR?

pyropy: You are right, I was thinking that it would automatically cache the request / response.

NikolasHaimerl: Thanks, is there a way to test this too?

wrangler.toml.example
NikolasHaimerl, line 1: Why is this named `.example`? Shouldnt this be where the production environment is defined?

bin/worker.js
```diff
@@ -1,5 +1,50 @@
...
+/**
+ * Parse query parameters from the request URL
+ * @param {Request} request
+ * @returns {{proofSetId: string; pieceCid: string; baseUrl: string;}}
+ */
+function parseSearchParams (request) {
+  const url = new URL(request.url)
+  const proofSetId = url.searchParams.get('proofSetId')
+  const pieceCid = url.searchParams.get('pieceCid')
+  const baseUrl = url.searchParams.get('baseUrl')
```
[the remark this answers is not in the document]
pyropy: Yeah subdomain makes sense, plus it's easy to setup so there shouldn't be much issues there. I've also seen @bajtos reference `https://download.filcdn.io/v1/<proofSet>/<pieceCID>` schema in Slack so I'm going to continue forward with it.

wrangler.toml.example
pyropy: I'm not exactly sure what was the motivation with that file. I have changed it so there's only `wranger.toml` now.

bin/worker.js
```diff
@@ -1,5 +1,23 @@
...
+    if (request.method !== 'GET') {
+      return new Response('Method Not Allowed', { status: 405 })
+    }
+
+    const { proofSetId, pieceCid } = parseRequest(request)
+    if (!proofSetId || !pieceCid) {
+      return new Response('Missing required fields', { status: 400 })
+    }
+
+    // TODO: Record retrieval stats to D1 asynchronously (do not block response)
```
juliangruber, line 20: https://developers.cloudflare.com/workers/runtime-apis/context/#waituntil

bin/worker.js
```diff
@@ -1,5 +1,23 @@
...
-    return await fetch(request)
+  async fetch (request, env, ctx, { retrieveFile = defaultRetrieveFile } = {}) {
+    if (request.method !== 'GET') {
+      return new Response('Method Not Allowed', { status: 405 })
+    }
+
+    const { proofSetId, pieceCid } = parseRequest(request)
+    if (!proofSetId || !pieceCid) {
+      return new Response('Missing required fields', { status: 400 })
+    }
```
juliangruber, line 18: would be nice to assert both fields individually, but for now, better to ship and iterate 👍

lib/retrieval.js
```diff
@@ -0,0 +1,21 @@
...
+ * @returns {Promise<Response>} - The response from the fetch request.
+ */
+export async function retrieveFile (baseUrl, pieceCid, cacheTtl = 86400) {
+  const url = `https://${baseUrl}/piece/${pieceCid}`
+  return fetch(url, {
+    cf: {
+      cacheTtlByStatus: {
+        '200-299': cacheTtl,
+        404: 0,
+        '500-599': 0
```
juliangruber, line 16: For the hackathon I propose to cache
- 2xx for a long time
- 4xx/5xx for no time

For production, we'll need a strategy for caching 4xx/5xx responses as well

pyropy: Yes, that was my plan. I was unsure of default TTL so I've set it to 24hrs but it's also configurable via env variables.

bin/worker.js
pyropy: I've created a separate issue for this #7

bin/worker.js
pyropy: @NikolasHaimerl I belive there's no actual way to do integration test on this but I've added tests that check if `fetch` has been called with cf caching params.

lib/request.js
```diff
@@ -0,0 +1,12 @@
+/**
+ * Parse params found in path of the request URL
+ *
+ * @param {Request} request
+ * @returns {{proofSetId?: string, pieceCid?: string}}
+ */
+export function parseRequest (request) {
+  const url = new URL(request.url)
+  const [proofSetId, pieceCid] = url.pathname.split('/').filter(Boolean)
```
juliangruber, line 9: will need to be updated once @NikolasHaimerl set up the new domains

juliangruber (approved):

NikolasHaimerl (approved):

lib/request.js
pyropy: Can you explain why? I'm not sure if domain name would affect this function unless we change how we construct the URL.

PR #13, opened 2025-05-23
feat: add d1 database

NikolasHaimerl:
This PR adds functionality to write retrieval logs to a d1 cloudflare database.


.github/workflows/ci.yml
```diff
@@ -10,35 +10,48 @@ jobs:
...
-          preCommands: |
-            cp wrangler.toml.example wrangler.toml
-            envsubst < wrangler.toml.example > wrangler.toml
-          postCommands: |
-            rm -f wrangler.toml
-          environment: production
-          secrets: |
-            INFLUX_TOKEN
+
+      - name: Install wrangler
```
pyropy, line 38: Can we keep `cloudflare/wrangler-action@v3` action instead of doing this manually?

I also belive that when using `cloudflare/wrangler-action@v3`  action you get wrangler CLI. We could do this part in the `preCommands` part of the wrangler action.

pyropy (commented):
Could you also take a look at cloudflare logs that (which Julian has previously shared https://developers.cloudflare.com/logs/). Maybe we could use that instead of the database to record and analyse the traffic?

.github/workflows/ci.yml
```diff
@@ -10,35 +10,48 @@ jobs:
...
         with:
           node-version: 22
-      - run: npm ci
+
+      - name: Install dependencies
+        run: npm ci
+
+      - name: Install Wrangler
+        run: npm install -g wrangler
+
```
juliangruber:
```suggestion
```

It should work without this as well, as long as we have `wrangler` in `devDependencies`

test/worker.test.js
```diff
@@ -16,3 +22,40 @@ describe('Cloudflare Worker', () => {
...
+    const HOSTNAME = 'example.com'
+    const PIECE_CID = 'QmExamplePieceCid1234'
+
+    const response = new Response(null, {
+      status: 200,
+      headers: { 'content-length': '1234' }
+    })
+
+    await logRetrievalResult(env, {
+      hostname: HOSTNAME,
```
juliangruber, line 37: this is missing the proofset id

juliangruber: since we want to store usage per proofset, not per hostname (a SP can have multiple proofsets)

migrations/0001_create_retrieval_logs_table.sql.sql
juliangruber, line 1: this file has a double sql ending

juliangruber, line 1: how is this migration going to be run in production?

juliangruber (changes requested):
please see Srdjan's comment and my reply

juliangruber:
> Could you also take a look at cloudflare logs that (which Julian has previously shared https://developers.cloudflare.com/logs/). Maybe we could use that instead of the database to record and analyse the traffic?

IIUC, we actually have to use that. The reason is caching: The worker won't be executed, if the response has already been cached. Therefore, we need to operate on the request logs level, and push to a data store from there

juliangruber:
EG:

1. Request to `/resource` -> cache miss -> worker will be executed
2. Request to `/resource` -> cache hit -> worker will _not_ be executed. In the current state of the PR, no request will be stored in D1

migrations/0001_create_retrieval_logs_table.sql.sql
NikolasHaimerl: by running `          wrangler d1 migrations apply filcdn-db
` at deployment.

migrations/0001_create_retrieval_logs_table.sql.sql
```diff
@@ -0,0 +1,10 @@
+-- Migration number: 0001 	 2025-05-23T08:31:24.810Z
+CREATE TABLE retrieval_logs (
+  id INTEGER PRIMARY KEY AUTOINCREMENT,
+  timestamp TEXT NOT NULL,
+  hostname TEXT NOT NULL,
+  piece_cid TEXT NOT NULL,
```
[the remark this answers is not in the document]
NikolasHaimerl: Addressed in https://github.com/CheckerNetwork/filcdn-worker/pull/13/commits/7f64973e2ab67b32420016f0e19519a3683c53a8

test/worker.test.js
NikolasHaimerl: Addressed in https://github.com/CheckerNetwork/filcdn-worker/pull/13/commits/7f64973e2ab67b32420016f0e19519a3683c53a8

migrations/0001_create_retrieval_logs_table.sql.sql
NikolasHaimerl: Addressed in https://github.com/CheckerNetwork/filcdn-worker/pull/13/commits/7f64973e2ab67b32420016f0e19519a3683c53a8

test/worker.test.js
```diff
@@ -16,3 +22,40 @@ describe('Cloudflare Worker', () => {
     expect(result).toBe(expected)
   })
 })
+
+describe('logRetrievalResult', () => {
```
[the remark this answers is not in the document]
NikolasHaimerl: Addressed in https://github.com/CheckerNetwork/filcdn-worker/pull/13/commits/0accb6ec673243844f05af0ca9d8b19c0643b7bf

migrations/0001_create_retrieval_logs_table.sql.sql
```diff
@@ -0,0 +1,10 @@
...
+  response_status INTEGER NOT NULL,
+  error_reason TEXT,
+  egress_bytes INTEGER NOT NULL
+);
```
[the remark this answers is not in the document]
NikolasHaimerl: Addressed in https://github.com/CheckerNetwork/filcdn-worker/pull/13/commits/0accb6ec673243844f05af0ca9d8b19c0643b7bf

migrations/0001_create_retrieval_logs_table.sql.sql
```diff
@@ -0,0 +1,10 @@
+-- Migration number: 0001 	 2025-05-23T08:31:24.810Z
+CREATE TABLE retrieval_logs (
+  id INTEGER PRIMARY KEY AUTOINCREMENT,
+  timestamp TEXT NOT NULL,
+  hostname TEXT NOT NULL,
+  piece_cid TEXT NOT NULL,
+  response_status INTEGER NOT NULL,
+  error_reason TEXT,
```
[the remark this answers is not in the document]
NikolasHaimerl: Addressed in https://github.com/CheckerNetwork/filcdn-worker/pull/13/commits/0accb6ec673243844f05af0ca9d8b19c0643b7bf

.github/workflows/ci.yml
NikolasHaimerl: It worked after some key iterations :D

juliangruber:
> 1. Do we care about data from cache hits? They have nothing to do with RSR scores of SPs and also don't require any egress from the SPs.

Yes, please read up on https://www.notion.so/spacemeridian/FilCDN-Overview-1f3cdd5cccdb80d9a935c9d2ade4059e?utm_content=1f3cdd5c-ccdb-80d9-a935-c9d2ade4059e&utm_campaign=T06QECYH087&n=slack&n=slack_link_unfurl&pvs=6. Pricing will be based both on CDN egress and on SP egress traffic. Just because the answer is cached, it doesn't mean it doesn't incur any cost for us.

> 2. Currently the cache is empty, so there should not be any cache hits, meaning the worker is executed as long as the request requires new files.

That is correct. But we also care about cache hits.

juliangruber:
> I think that @pyropy implemented caching in his worker in the https://github.com/CheckerNetwork/filcdn-worker/pull/8 for retrievals, correct? That would mean that cach miss or hits are also recorded in the database.

Thank you for bringing this up! I hadn't thought about this properly.

There are two relevant ways you can set up caching in Cloudflare workers:

1. Caching inside the worker
2. Caching outside of the worker

Caching **inside** the worker is what Srdjan did in his PR. It means that for every incoming CDN request, the worker is executed. The worker's request to the SP can be cached by Cloudflare. In this case, correct, we can submit logs to D1 from inside the worker.

Caching **outside** the worker works by setting cache headers on the HTTP response produced from inside the worker. That means that on a cached response, the worker is _not_ executed.

See https://developers.cloudflare.com/workers/examples/cache-using-fetch/ for both caching methods being used in one example.

Caching outside the worker is faster for the end user. A question is whether it's more work for us or not. For the Hackathon, we should be able to ship with the current caching strategy. We eventually 100% should switch to caching outside the worker, if possible.

@bajtos wdyt?

pyropy:
> Caching **outside** the worker works by setting cache headers on the HTTP response produced from inside the worker. That means that on a cached response, the worker is _not_ executed.

By that I suppose you reference to setting the `Cache-Control` headers? In that case caching is the one that's doing the caching and will not send the outbound request, hence we cannot measure it.

.github/workflows/ci.yml
```diff
@@ -24,21 +27,22 @@ jobs:
...
-          accountId: ${{ secrets.CLOUDFLARE_ACCOUNT_ID }}
-          preCommands: |
-            cp wrangler.toml.example wrangler.toml
-            envsubst < wrangler.toml.example > wrangler.toml
-          postCommands: |
-            rm -f wrangler.toml
-          environment: production
-          secrets: |
-            INFLUX_TOKEN
+      - uses: cloudflare/wrangler-action@v3
```
pyropy, line 30: This part actually replaces the whole deployment step with database creaiton.

test/worker.test.js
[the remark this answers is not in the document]
pyropy: I don't see that file in https://github.com/CheckerNetwork/filcdn-worker/commit/0accb6ec673243844f05af0ca9d8b19c0643b7bf commi 👀

lib/store.js
```diff
@@ -0,0 +1,32 @@
+/**
+ * Logs the result of a file retrieval attempt to the D1 database.
+ *
+ * @param {object} env - Worker environment (contains D1 binding).
+ * @param {object} context - Retrieval context: { hostname, pieceCid, response, error }
+ * @returns {Promise<void>} - A promise that resolves when the log is inserted.
+ */
+export async function logRetrievalResult (env, { hostname, pieceCid, response, cacheMiss, proofSetId }) {
```
bajtos, line 8: I don't see this function being called. I assume you will integrate this function into the retrieval workflow later, after this PR is landed, is that correct?

juliangruber:
> > Caching **outside** the worker works by setting cache headers on the HTTP response produced from inside the worker. That means that on a cached response, the worker is _not_ executed.
> 
> By that I suppose you reference to setting the `Cache-Control` headers?

Yep

> In that case caching is the one that's doing the caching and will not send the outbound request, hence we cannot measure it.

In both cases Cloudflare performs the caching. The difference is whether the whole worker execution is cached, or part of it.

It would be cleaner and faster to cache the whole worker execution, but for the MVP we can do what's simplest and perform caching inside the worker

pyropy:
> > > Caching **outside** the worker works by setting cache headers on the HTTP response produced from inside the worker. That means that on a cached response, the worker is _not_ executed.
> > 
> > 
> > By that I suppose you reference to setting the `Cache-Control` headers?
> 
> Yep
> 
> > In that case caching is the one that's doing the caching and will not send the outbound request, hence we cannot measure it.
> 
> In both cases Cloudflare performs the caching. The difference is whether the whole worker execution is cached, or part of it.
> 
> It would be cleaner and faster to cache the whole worker execution, but for the MVP we can do what's simplest and perform caching inside the worker

I belive this is trivial to add, we can do it right away with some sane TTL default.

lib/store.js
NikolasHaimerl: Yes that is correct, Srdjan and I are working in parallel so at the time of creating this PR there was no way of integrating it with the worker.

test/setup-db.js
```diff
@@ -0,0 +1,22 @@
...
+    import: 'default',
+    eager: true
+  })
+
+  const migrations = Object.entries(modules)
+    .sort(([a], [b]) => a.localeCompare(b))
+    .map(([, sql]) => sql)
+
+  for (const sql of migrations) {
+    const statements = sql
```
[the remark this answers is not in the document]
NikolasHaimerl: I don't quite understand what you mean. This setup is for the test instance which is running in memory and thus is deleted right after it is executed.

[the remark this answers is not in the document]
bajtos: I see, I didn't realise this is used by tests only.

How do you envision applying the migration scripts on the production D1 instance?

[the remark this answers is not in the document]
bajtos: (This thread is not blocking this PR from being landed. )

bajtos:
Re: caching outside or inside the worker.

I don't know enough about Cloudflare workers to tell which option we need.

What's important to me: We need to collect stats about the egress served by our CDN, including responses serving files from the cache.
- If the caching strategy used by the retrieval worker does not allow us to do that, then we need to have a way to achieve that. E.g. by adding another lightweight worker that's executed for *all* requests, logging the stats, and falling back to Cloudflare cache + retrieval worker to do the retrieval.
- OTOH, two workers seems enough of a hassle that it's worth choosing the strategy that will give us stats about responses served from the cache in the retrieval worker.

migrations/0001_create_retrieval_logs_table.sql.sql
bajtos: Can you document that step in the README please (if you haven't already).

Related thread: https://github.com/CheckerNetwork/filcdn-worker/pull/13#discussion_r2104600641

NikolasHaimerl: It is part of the workflow ci, do you want to additionally document me in the README? Should we document the entire workflow then for consistency?

test/worker.test.js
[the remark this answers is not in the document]
NikolasHaimerl: Sorry it is this commit https://github.com/CheckerNetwork/filcdn-worker/commit/7f64973e2ab67b32420016f0e19519a3683c53a8

lib/store.js
```diff
@@ -0,0 +1,32 @@
...
+  const timestamp = new Date().toISOString()
+  const responseStatus = response.status
+  const contentLength = response.headers.get('content-length')
+  const egressBytes = contentLength ? parseInt(contentLength, 10) : 0
+
+  try {
+    await env.DB.prepare(`
+        INSERT INTO retrieval_logs 
+          (timestamp, hostname, piece_cid, response_status, egress_bytes, cache_miss, proof_set_id)
+        VALUES (?, ?, ?, ?, ?, ?, ?)
```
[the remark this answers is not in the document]
NikolasHaimerl: We could also do a combination of rolling stats (rsr scores,..) and timestamp based data which is only kept for a certain period of time (depending on how much data is created per time period)

migrations/0001_create_retrieval_logs_table.sql
```diff
@@ -0,0 +1,11 @@
+-- Migration number: 0001 	 2025-05-23T08:31:24.810Z
+CREATE TABLE retrieval_logs (
+  id INTEGER PRIMARY KEY AUTOINCREMENT,
+  timestamp TEXT NOT NULL,
+  hostname TEXT NOT NULL,
```
NikolasHaimerl, line 5: I think we can remove this field, piece_cid and proof_set_id should be sufficient and we can then safe some bytes from being written. WDYT @bajtos

lib/store.js
```diff
@@ -0,0 +1,32 @@
...
+ *
+ * @param {object} env - Worker environment (contains D1 binding).
+ * @param {object} context - Retrieval context: { hostname, pieceCid, response, error }
+ * @returns {Promise<void>} - A promise that resolves when the log is inserted.
+ */
+export async function logRetrievalResult (env, { hostname, pieceCid, response, cacheMiss, proofSetId }) {
+  const timestamp = new Date().toISOString()
+  const responseStatus = response.status
+  const contentLength = response.headers.get('content-length')
+  const egressBytes = contentLength ? parseInt(contentLength, 10) : 0
```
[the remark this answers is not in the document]
bajtos: > Counting the bytes of the return body could be done like so:

I think you also need to forward the data back to the client who made the request? But yes, I think your approach works at the high level 👍🏻 

---

**It seems that Curio is returning Content-Length, therefore we can rely on it for the June 6 hackathon.**

I propose to log a message when this assumption is not met, so that we have a way how to verify whether it's always the case.


```js
if (contentLength === undefined) {
  // log a message
  // include identifiers allowing us to reproduce this
}
const egressBytes = contentLength ? parseInt(contentLength, 10) : 0
```

Here is an example request I used to check if the SP is reporting the content length:

```
❯ curl -i https://yablu.net/piece/baga6ea4seaqofzdf7xcxihwvgicg3o7bqei5omukufmmud62eutbzti6v2ntahq
HTTP/2 200
accept-ranges: bytes
access-control-allow-headers: Content-Type, Authorization, Accept, Accept-Encoding
access-control-allow-methods: GET, POST, PUT, DELETE, OPTIONS
access-control-allow-origin: *
cache-control: public, max-age=29030400, immutable
content-security-policy: default-src 'self'
content-type: audio/wave
etag: baga6ea4seaqofzdf7xcxihwvgicg3o7bqei5omukufmmud62eutbzti6v2ntahq
last-modified: Thu, 01 Jan 1970 00:00:00 GMT
referrer-policy: strict-origin-when-cross-origin
strict-transport-security: max-age=63072000; includeSubDomains
vary: Accept-Encoding
x-content-type-options: nosniff
x-frame-options: DENY
x-xss-protection: 1; mode=block
content-length: 158257942
date: Mon, 26 May 2025 12:12:38 GMT
```

lib/store.js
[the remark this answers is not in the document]
NikolasHaimerl: I am closing this thread as we discussed that for the 6.June MVP we will not implement any scaling optimization just yet.

NikolasHaimerl:
> > > > > Caching **outside** the worker works by setting cache headers on the HTTP response produced from inside the worker. That means that on a cached response, the worker is _not_ executed.
> > > > 
> > > > 
> > > > By that I suppose you reference to setting the `Cache-Control` headers?
> > > 
> > > 
> > > Yep
> > > > In that case caching is the one that's doing the caching and will not send the outbound request, hence we cannot measure it.
> > > 
> > > 
> > > In both cases Cloudflare performs the caching. The difference is whether the whole worker execution is cached, or part of it.
> > > It would be cleaner and faster to cache the whole worker execution, but for the MVP we can do what's simplest and perform caching inside the worker
> > 
> > 
> > I belive this is trivial to add, we can do it right away with some sane TTL default.
> 
> @juliangruber Here's the PR for caching outside of the CF worker #15

I am resolving this thread, as we discussed that for the MVP we will have all requests go through our worker.

wrangler.toml
```diff
@@ -9,4 +9,9 @@ CACHE_TTL = 86400
 
 [env.production.vars]
 ENVIRONMENT = "production"
+
+[[d1_databases]]
+binding = "DB"
+database_name = "filcdn-stats"
+database_id = "filcdn-stats"
 CACHE_TTL = "$CACHE_TTL"
```
pyropy, line 17:
```suggestion
[env.production.vars]
ENVIRONMENT = "production"
CACHE_TTL = "$CACHE_TTL"

[[d1_databases]]
binding = "DB"
database_name = "filcdn-stats"
database_id = "filcdn-stats"
```

migrations/0001_create_retrieval_logs_table.sql
pyropy, line 5: Should we add `client_address` column now before we've added the `/client/commp` URL structure? WDYT?

Also we could swap out `hostname` with `proof_set_owner` or something similar?

pyropy (approved):
I think this looks good for the current iterration where we have `/<proof-set-id>/<commp>` URL structure. 

We might create a follow-up PRs with new migration that removes / adds some columns.

migrations/0001_create_retrieval_logs_table.sql
NikolasHaimerl: I think `client_addess` makes sense now that we move to a different URL structure. `hostname` is supposed to be the URL of the SP, I am not sure whether `proof_set_owner` really targets the right data that should be captured.

lib/store.js
```diff
@@ -0,0 +1,37 @@
...
+ * @param {object} env - Worker environment (contains D1 binding).
+ * @param {object} context - Retrieval context: { hostname, pieceCid, response, error }
+ * @returns {Promise<void>} - A promise that resolves when the log is inserted.
+ */
+export async function logRetrievalResult (env, { hostname, pieceCid, response, cacheMiss, proofSetId }) {
+  const timestamp = new Date().toISOString()
+  const responseStatus = response.status
+  const contentLength = response.headers.get('content-length')
+  if (!contentLength) {
+    debug('No content-length header found in response for pieceCid: and hostname:', pieceCid, hostname)
```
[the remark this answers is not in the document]
juliangruber: Otherwise this will log:

```
No content-length header found in response for pieceCid: and hostname: <pieceCid> <hostname>
```

.github/workflows/ci.yml
```diff
@@ -30,6 +32,11 @@ jobs:
           accountId: ${{ secrets.CLOUDFLARE_ACCOUNT_ID }}
           preCommands: |
             envsubst < wrangler.toml > wrangler.toml.tmp && mv wrangler.toml.tmp wrangler.toml
+           &&
+            if ! wrangler d1 db list | grep -q filcdn-db; then
+              wrangler d1 db create filcdn-db
```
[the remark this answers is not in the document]
juliangruber: is this missing `--local`?

juliangruber (approved):

PR #15, opened 2025-05-23
add: Cache response outside of CF worker

pyropy:
This pull request introduces caching response outside of CF worker by setting `Cache-Control` headers on response. Cache is performed only on successful response.

### DIFF
--- a/bin/worker.js
+++ b/bin/worker.js
@@ -18,10 +18,12 @@ export default {
     }
 
     // TODO: Record retrieval stats to D1 asynchronously (do not block response)
-    const response = await retrieveFile(BASE_URL, pieceCid, env.CACHE_TTL)
+    let response = await retrieveFile(BASE_URL, pieceCid, env.CACHE_TTL)
     if (response.ok) {
+      response = new Response(response.body, response)
       response.headers.set('Cache-Control', `max-age=${env.CACHE_TTL}`)
     }
+
     return response
   }
 }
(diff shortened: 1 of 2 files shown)


bin/worker.js
```diff
@@ -18,6 +18,12 @@ export default {
     }
 
     // TODO: Record retrieval stats to D1 asynchronously (do not block response)
-    return await retrieveFile(BASE_URL, pieceCid, env.CACHE_TTL)
+    let response = await retrieveFile(BASE_URL, pieceCid, env.CACHE_TTL)
+    if (response.ok) {
+      response = new Response(response.body, response)
+      response.headers.set('Cache-Control', `max-age=${env.CACHE_TTL}`)
+    }
```
juliangruber, line 25:
```suggestion
    const response = await retrieveFile(BASE_URL, pieceCid, env.CACHE_TTL)
    if (response.ok) {
      response.headers.set('Cache-Control', `max-age=${env.CACHE_TTL}`)
    }
```

Would this work as well?

pyropy: @juliangruber That will not work as response headers are immutable.

juliangruber (changes requested):
We decided not to do this for now, so that we can log req/res details for every request

bajtos:
> We decided not to do this for now, so that we can log req/res details for every request

💯

bajtos:
Let's close this pull request for now. We can re-open it later if we decide to add outside caching again.

PR #21, opened 2025-05-28
feat!: subdomain handling

bajtos:
Change the request URL format to

    https://{wallet}.calibration.filcdn.io/{CommP}


lib/request.js
```diff
@@ -2,11 +2,24 @@
  * Parse params found in path of the request URL
  *
  * @param {Request} request
- * @returns {{proofSetId?: string, pieceCid?: string}}
+ * @returns {{clientWalletAddress?: string, pieceCid?: string}}
```
pyropy, line 5:
```suggestion
 * @returns {{clientWalletAddress?: string, pieceCid?: string, error?: string}}
```

There's `error` field missing in the return type.

pyropy (approved):
Looks good to me. Let's just add the missing `error` field to the return type before we merge.

PR #32, opened 2025-05-29
refactor: move CACHE_TTL config to git

bajtos:
Move the TTL configuration from GitHub-managed environmen variables to wrangler.toml versioned in git. This way we can keep track of changes made to this configuration.


pyropy (approved):
🚀

PR #34, opened 2025-05-29
fix: Incorrect database ID

pyropy:
- Set correct database id to wranger.toml file


bajtos (approved):
Nice! I don't know how to check if this is the correct ID to use, but I don't see any obvious problems.

Issue #28, opened 2025-05-28
Collect & report metrics to FilOz


bajtos:
See https://www.notion.so/filecoindev/Collect-service-metrics-and-stats-200dc41950c180c98b28f2d2445ffe6f

Collect in the worker:

- [x] #/rate of CDN requests over time
- [x] Egress served by CDN over time
- per-SP stats
  - [x] # of requests (cache-miss vs total requests)
  - [x] egress (cache-miss vs total)
  - [x] TTFB of fetching from SP on cache-miss
  - [x] retrieval speed (avg, p95) → stretch goal
- [x] Geo region of the clients hitting on CDN
- [x] CDN cache hit vs miss rate

Visualise on https://dashboard.filcdn.com:

- [x] #/rate of CDN requests over time
- [x] Egress served by CDN over time
- per-SP stats
  - [x] # of requests (cache-miss vs total requests)
  - [x] egress (cache-miss vs total)
  - [x] TTFB of fetching from SP on cache-miss
  - [x] retrieval speed (avg, p95) → stretch goal
- [x] Geo region of the clients hitting on CDN
- [x] CDN cache hit vs miss rate

pyropy:
> egress (cache-miss vs total)

I would argue that cache-miss egress is the total egress of the SP as they wont have any egress on cache-hit.

pyropy:
> Geo region of the clients hitting on CDN

👉🏻  https://developers.cloudflare.com/fundamentals/reference/http-headers/#cf-ipcountry

PR #35, opened 2025-05-29
feat: add performance stats

NikolasHaimerl:
This PR adds two performance stats:
- Time to first byte (ttfb)  for fetching the data from cache/SP
- Time to first byte since the worker execution started

Links:
- https://github.com/filcdn/worker/issues/28


bin/worker.js
```diff
@@ -17,6 +17,8 @@ export default {
    * @returns
    */
   async fetch(request, env, ctx, { retrieveFile = defaultRetrieveFile } = {}) {
+    // Worker entry timestamps
+    const t0 = performance.now()
```
bajtos, line 21: Let's use a more descriptive name, please. E.g. 

```suggestion
    const startedAt = performance.now()
```

bin/worker.js
```diff
@@ -31,11 +33,14 @@ export default {
       return new Response('Missing required fields', { status: 400 })
     }
 
+    // Timestamp to measure file retrieval performance
+    const t1 = performance.now()
```
bajtos, line 37: I think we should measure the time for error responses too, shouldn't we?

(See `return new Response('Missing required fields', { status: 400 })` on the line above - when we return from the handler early, timing is not recorded.)

bin/worker.js
```diff
@@ -31,11 +33,14 @@ export default {
...
     const { response, cacheMiss, contentLength } = await retrieveFile(
       BASE_URL,
       pieceCid,
       env.CACHE_TTL,
     )
+    const ttfb = performance.now() - t1
```
bajtos, line 43: I think we should measure TTFB from the time our handler (worker) started. That's closer to the TTFB observed by end users.

bin/worker.js
```diff
@@ -44,6 +49,10 @@ export default {
         contentLength,
         responseStatus: response.status,
         timestamp: new Date().toISOString(),
+        performanceStats: {
+          ttfb,
+          workerExecutionTime: performance.now() - t0,
```
bajtos, line 54: This will use `performance.now()` at the time we call `logRetrievalResult`. I.e. it will be almost the exact same value as what `performance.now()` returns in the line calculating TTFB.

I find this confusing to understand and reason about.

How can we make it easier to reason about these two values?

My perspective:

As a CDN user, I am interested in two values
- TTFB - time to first byte, how quickly can I start watching the video?
- TTLB - time to last byte, how quickly can I download the entire file?

We obviously cannot measure timing info from the client side. I would like us to collect measurements that are as close to the client metrics as possible.

That means ttfb is should be based on `t0` (not `t1`)

I am not sure what is the value of `workerExecutionTime`, I suppose the current solution is a good start. It would be great to cross-check the values reported in workerExecutionTime with the timing info collected by Cloudflare. I am concerned that our timing will be shorter because we don't wait for the worker to stream the entire response.

bajtos (changes requested):

bin/worker.js
pyropy: +1 on more descriptive naming

bin/worker.js
NikolasHaimerl: The reasoning of @pyropy and I for t0 and t1 is that in the future as soon as @juliangruber pushes his event listener there will be quite a few read and write operations performed on various databases. These should not be part of the ttfb as ttfb measures the time to first byte between filCDN and the SP. It should however be captured by worker_execution_time as this metric measures the time it takes from the client's request reaching the CDN to the point where the response leaves the CDN.

bajtos: I see. I assumed TTFB is the time to first byte after the request arrived to our handler.

I agree it makes sense to measure TTFB for cache-miss requests to SPs.

Let's use a more descriptive name please, to make this difference clear. For example `cache_miss_ttfb` or `sp_ttfb`.

How do you differentiate between the case where we fetch the data from Cloudflare cache vs. cache-miss to the SP? We shouldn't be recording TTFB in the first case.

NikolasHaimerl: The database records whether it was a cache miss or not.

bin/worker.js
bajtos: I'd like to see the number of all requests on our dashboard, not only the number of "valid" requests.

I created a follow-up issue for this: https://github.com/filcdn/roadmap/issues/272

bajtos:
I think this PR is good to be landed. @pyropy PTAL.

lib/store.js
```diff
@@ -50,6 +54,8 @@ export async function logRetrievalResult(
         responseStatus,
         egressBytes,
         cacheMiss,
+        performanceStats?.fetchTtfb || null, // Time to first byte, if available
+        performanceStats?.workerTtfb || null, // Worker execution time, if available
```
pyropy, line 58:
```suggestion
        performanceStats?.fetchTtfb || null,
        performanceStats?.workerTtfb || null,
```

pyropy (approved):
I think this is ready to ship 🚢  I have one minor change request regarding the code comments but other than that I think we're good.

lib/store.js
bajtos: See 6b51f2e

pyropy (approved):

PR #47, opened 2025-05-30
fix: --remote for DB migrations executed in CD

bajtos:
Fix the problem where the CD pipeline did not apply DB migrations on the remote database.


pyropy (approved):
🚢

PR #48, opened 2025-05-30
test: fix flaky test assertion

bajtos:
Our code runs so fast that TTFB values are often `0`.

This commit changes the test assertions to accept such value.


pyropy (approved):
:shipit:

Issue #54, opened 2025-06-02
Index PDPVerifier and Pandora service events


bajtos:
To enable the retrieval worker to quickly verify the validity of the retrieval request and map the CommP to the retrieval URL, we must maintain a view of the subset of PDPVerifier and Pandora state in the Cloudflare data layer (e.g., D2). 

This is blocking the following tasks:
- #41 
- #42 
- #43 
- #44 

In those tasks, we have the following inputs:

- CDN client address (`0x...`)
- CommP (`bafy...`)

We need to answer the following questions:
- Is CommP included in client's deals that are paying for CDN retrievals?
- Which SPs store the CommP and what are their PDP owner addresses?

It would be great if the index allowed us to discriminate between the following two cases:
- The client does not have any PDP deal for the provided CommP. (The CDN will return a 4xx code, e.g. 400 Bad Request or 404 Not Found.)
- The client has a PDP deal for the provided CommP, but the deal does not include CDN payment. (The CDN will return [402 Payment Required](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status/402).)

Why is this important: we want to make it clear whether the combination of Client+CommP is not valid (e.g. because we didn't index the recently created PDP deal yet) or whether the client should pay for the CDN service (we want to nudge them to upgrade!).

bajtos:
Closing as done.

PR #56, opened 2025-06-02
chore: switch to monorepo setup + introduce calibration env

bajtos:
Introduce two workers in one monorepo, sharing the same database:

- indexer is watching PDPVerifier and Pandora events
- retriever is handling CDN retrieval requests

The DB schema is owned by the retriever worker, I could not find an easy way to have a monorepo-level migrations.

As part of this change, I changed the environment name from "production" to "calibration", to match the fact that our current deployment is linked to the Filecoin Callibration testnet.

See:
- https://github.com/filcdn/worker/issues/54

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -34,7 +34,7 @@ jobs:
           apiToken: ${{ secrets.CLOUDFLARE_API_TOKEN }}
           accountId: ${{ secrets.CLOUDFLARE_ACCOUNT_ID }}
           preCommands: './bin/pre-deploy.sh'
-          environment: calibnet
+          environment: calibration
       - name: Deploy Indexer
         uses: cloudflare/wrangler-action@v3
         with:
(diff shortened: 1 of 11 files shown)


pyropy (changes requested):
Are workers supposed to have one migrations directory or both workers would have separate migration directories?

I'm not exactly sure how do worker migrations work, but if latter is the case I think it would be a bad idea to use monorepo structure. 

In that case I think it would make sense to either:

- Create a third workspace used to host migrations that they both can import
- Simplify the repository from monorepo to structure where code is shared in `lib` but there is multiple different `bin` entries. (Note that this might only work if wranger.toml supports multiple worker definitions).

juliangruber:
@bajtos Typescript is passing for me now:

```
$ npm run lint

> @filcdn/monorepo@1.0.0 lint
> eslint && prettier --check . && tsc -p .

Checking formatting...
All matched files use Prettier code style!
$ echo $?
0
```

juliangruber:
Suspicion: Configuring dbs in multiple wrangler configs leads to different dbs on disk, even if they share the same id

juliangruber:
Tests pass now locally, the path to the migrations folder was wrong

juliangruber:
@bajtos I have no appetite to look further into type safety for the tests. I find the monorepo setup more important than types, at this stage. I hope we can eventually figure them out.

I'll now create the indexer PR, and use branch `monorepo` as the target

juliangruber (approved):

PR #60, opened 2025-06-03
add: Time to last byte

pyropy:
This PR introduces "fetch time to last byte" measurement. This measurement is meant to be used to calculate retrieval speed (file size / time to last byte). 

Related to #28


bajtos (changes requested):
Nice work! Let's wait until the other PR adding `measureStreamedEgress` is merged first.

retriever/bin/retriever.js
```diff
@@ -87,12 +89,14 @@ export default {
     ctx.waitUntil(
       (async () => {
         const egressBytes = await measureStreamedEgress(reader)
+        const lastByteAt = performance.now()
```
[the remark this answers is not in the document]
pyropy: Addressed in https://github.com/filcdn/worker/pull/60/commits/f8e3edf8dbc70e7d1c68c3f5a625fe5aa707e40d

retriever/bin/retriever.js
```diff
@@ -79,6 +80,7 @@ export default {
           egressBytes: 0, // No body to measure
           performanceStats: {
             fetchTtfb: firstByteAt - fetchStartedAt,
+            fetchTtlb: firstByteAt - fetchStartedAt,
```
NikolasHaimerl, line 83: Why are we measuring the same stat twice?

pyropy: In the case of no response TTFB and TTLB is the same as there is no response to be read. In case where there is a response body TTLB (Time-to-last-byte) is captured once we've finished reading the whole body

```js
const egressBytes = await measureStreamedEgress(reader)
const lastByteFetchedAt = performance.now()
```

In the end TTLB is calculated as `lastByteFetchedAt - fetchStartedAt` as you can see in the performance stats object:

```js
performanceStats: {
  fetchTtfb: firstByteAt - fetchStartedAt,
  fetchTtlb: lastByteFetchedAt - fetchStartedAt,
  workerTtfb: firstByteAt - workerStartedAt,
}
```

bajtos (approved):
LGTM 👍🏻

PR #62, opened 2025-06-03
Add index worker implementation

juliangruber:
This ports the filcdn index service over to the worker architecture (powered by goldsky).

<del>Next: Implement new events</del>

### DIFF
--- a/indexer/package.json
+++ b/indexer/package.json
@@ -10,19 +10,5 @@
     "deploy:calibration": "wrangler deploy --env calibration",
     "start": "wrangler d1 migrations apply dev-db --local --env dev --cwd ../db  && wrangler dev --env dev",
     "test": "vitest run"
-  },
-  "prettier": "@checkernetwork/prettier-config",
-  "dependencies": {
-    "debug": "^4.4.1",
-    "neostandard": "^0.12.1",
-    "prettier": "^3.5.3"
-  },
-  "devDependencies": {
-    "@checkernetwork/prettier-config": "^1.0.1",
-    "@cloudflare/vitest-pool-workers": "^0.8.34",
-    "@types/node": "^22.15.29",
-    "typescript": "^5.8.3",
-    "vitest": "3.1.4",
-    "wrangler": "^4.17.0"
   }
 }
--- a/package-lock.json
+++ b/package-lock.json
@@ -24,20 +24,7 @@
     },
     "indexer": {
       "name": "@filcdn/indexer",
-      "version": "1.0.0",
-      "dependencies": {
-        "debug": "^4.4.1",
-        "neostandard": "^0.12.1",
-        "prettier": "^3.5.3"
-      },
-      "devDependencies": {
-        "@checkernetwork/prettier-config": "^1.0.1",
-        "@cloudflare/vitest-pool-workers": "^0.8.34",
-        "@types/node": "^22.15.29",
-        "typescript": "^5.8.3",
-        "vitest": "3.1.4",
-        "wrangler": "^4.17.0"
-      }
+      "version": "1.0.0"
     },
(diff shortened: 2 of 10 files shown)


indexer/bin/indexer.js
```diff
@@ -6,6 +6,93 @@ export default {
...
+    if (request.method !== 'POST') {
+      return new Response('Method Not Allowed', { status: 405 })
+    }
+    const payload = await request.json()
+    const pathname = new URL(request.url).pathname
+    if (pathname === '/proof-set-created') {
+      if (!payload.set_id || !payload.owner) {
+        return new Response('Bad Request', { status: 400 })
+      }
+      await env.DB.prepare(
```
pyropy, line 18: I think we should move request handling logic into handlers similar to what we did in projects like `spark-api`. I think it'll make a great difference when maintaining this project as this already gets a bit tought to read.

pyropy (changes requested):

indexer/bin/indexer.js
juliangruber: Let's not worry about maintainability for now, this can be refactored post launch

db/migrations/0004_add_indexer.sql
```diff
@@ -0,0 +1,18 @@
+CREATE TABLE IF NOT EXISTS indexer_proof_sets (
+  set_id TEXT PRIMARY KEY,
```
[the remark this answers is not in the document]
juliangruber: Which type would you suggest?

indexer/bin/indexer.js
bajtos: I was thinking about the same - we should extract route handlers into individual functions, and it's a refactoring to do post M1.

db/migrations/0004_add_indexer.sql
```diff
@@ -0,0 +1,27 @@
...
+-- E.g. the value 0x0...01 is stored as "1"
+
+CREATE TABLE IF NOT EXISTS indexer_proof_sets (
+  set_id TEXT PRIMARY KEY,
+  owner TEXT NOT NULL
+);
+
+CREATE TABLE IF NOT EXISTS indexer_roots (
+  root_id TEXT NOT NULL,
+  set_id TEXT NOT NULL,
```
NikolasHaimerl, line 14: Do we want to mark this with a foreign key?

indexer/bin/indexer.js
```diff
@@ -6,6 +6,99 @@ export default {
...
+          VALUES (?, ?, ?, ?, ?)
+          ON CONFLICT DO NOTHING
+        `,
+      )
+        .bind(
+          payload.proof_set_id,
+          payload.rail_id,
+          payload.payer,
+          payload.payee,
+          payload.with_cdn || null,
```
NikolasHaimerl, line 96: Is null itended or should we set it to false?

NikolasHaimerl (approved):
I have two non blocking comments. Feel free to merge if not applicable

db/migrations/0004_add_indexer.sql
bajtos: Great question!

My gut feeling is that the events can in theory arrive out of order, e.g. first a new root, second a new proof set, in which case a foreign key would prevent us from recording the new root.

bajtos (approved):
The current version is good enough as far as I am concerned.

db/migrations/0004_add_indexer.sql
juliangruber: I had it at first like this as well, but it didn't work, since events can arrive out of order. The set this is referencing might not be known to us yet

indexer/bin/indexer.js
juliangruber: null is intended, I also want to know whether this property existed in the payload

PR #65, opened 2025-06-04
Add auth to indexer

juliangruber:
Screenshot 2025-06-04 at 08 29 38

Screenshot 2025-06-04 at 08 29 44

@bajtos I believe that is what you found as well. Do you know how to at least disable this ts failure?

### DIFF
--- a/indexer/test/indexer.test.js
+++ b/indexer/test/indexer.test.js
@@ -4,6 +4,9 @@ import { env } from 'cloudflare:test'
 
 const randomId = () => String(Math.ceil(Math.random() * 1e10))
 
+env.SECRET_HEADER_KEY = 'secret-header-key'
+env.SECRET_HEADER_VALUE = 'secret-header-value'
+
 describe('retriever.indexer', () => {
   it('requires authentication', async () => {
     const req = new Request('https://host/', { method: 'POST' })


bajtos:
>I believe that is what you found as well. Do you know how to at least disable this ts failure?

First of all, after you add new env vars or secrets, we need to run `npx wrangler types --cwd indexer` to update the TypeScript interfaces. 

For secrets, you need to add them to `index/.dev.env` first, so that `wrangler types` picks them up.

As for the error, I am encountering it too. I think we need to change TypeScript setup to make each worker (indexer, retriever) an isolated project.

For now, I propose to ignore the error. This is what I have in local code:

```ts
  async fetch(
    request,
    env,
    ctx,
    { createPdpVerifierClient = defaultCreatePdpVerifierClient },
  ) {
    // TypeScript setup is broken in our monorepo
    // There are multiple global Env interfaces defined (one per worker),
    // TypeScript merges them in a way that breaks our code.
    // We should eventually fix that.
    // @ts-ignore
    const { GLIF_TOKEN, RPC_URL, PDP_VERIFIER_ADDRESS } = env
```

pyropy (approved):
Looks good to me. Let's solve the ts type problem and ship this 🚢

indexer/bin/indexer.js
```diff
@@ -6,6 +6,12 @@ export default {
    * @returns {Promise<Response>}
    */
   async fetch(request, env, ctx) {
+    const secretHeaderKey = env.SECRET_HEADER_KEY
+    const secretHeaderValue = env.SECRET_HEADER_VALUE
```
bajtos, line 10: Would this fix the problem?

```suggestion
    // TypeScript setup is broken in our monorepo
    // There are multiple global Env interfaces defined (one per worker),
    // TypeScript merges them in a way that breaks our code.
    // We should eventually fix that.
    // @ts-ignore
    const { SECRET_HEADER_KEY, SECRET_HEADER_VALUE } = env
```

indexer/test/indexer.test.js
```diff
@@ -250,6 +294,9 @@ describe('retriever.indexer', () => {
       const railId = randomId()
       const req = new Request('https://host/proof-set-rail-created', {
         method: 'POST',
+        headers: {
+          [env.SECRET_HEADER_KEY]: env.SECRET_HEADER_VALUE,
+        },
```
bajtos, line 299: Is it time to extract a helper function so that we don't have to repeat this header configuration in every test, and we don't have to update every test if we introduce more headers in the future?

```js
async function invokeWebhook(endpointPath, body) {
  const req = new Request(`https://host${endpointPath}`, {
    method: 'POST',
    headers: {
      [env.SECRET_HEADER_KEY]: env.SECRET_HEADER_VALUE,
    },
    body: JSON.stringify(body)
  )
  return fetch(req)
}
```

indexer/bin/indexer.js
```diff
@@ -6,6 +6,12 @@ export default {
    * @returns {Promise<Response>}
    */
   async fetch(request, env, ctx) {
+    const secretHeaderKey = env.SECRET_HEADER_KEY
```
bajtos, line 9: Why are we treating the secret header key (header name) as a sensitive secret? IMO, we can hard-code this value inside `indexer.js` in plain text.

I mean, like, most web services use the well-specified `Authorization` header name, there is no secrecy around that name.

bajtos (approved):
Looks good to me, thanks for including tests for this new code!

indexer/bin/indexer.js
juliangruber: This is the name in goldsky: "Secret header key". I don't care too much about whether it's a secret or not. This is already implemented.

indexer/test/indexer.test.js
juliangruber: I don't see a need for refactors at this point

juliangruber:
> > I believe that is what you found as well. Do you know how to at least disable this ts failure?
> 
> First of all, after you add new env vars or secrets, we need to run `npx wrangler types --cwd indexer` to update the TypeScript interfaces.
> 
> For secrets, you need to add them to `index/.dev.env` first, so that `wrangler types` picks them up.

This has already been done in the PR.

PR #73, opened 2025-06-05
feat: log getRootCid() failures

bajtos:


indexer/bin/indexer.js
```diff
@@ -88,12 +88,27 @@ export default {
...
+       */
+      const maybeGetRootCid = async (setId, rootId) => {
+        try {
+          const cid = await pdpVerifier.getRootCid(setId, BigInt(rootId))
+          return cid
+        } catch (/** @type {any} */ err) {
+          console.error(
+            `Cannot get root CID for setId=${setId} rootId=${rootId}: ${err?.stack ?? err}`,
+          )
+          return null
```
juliangruber, line 104: I think we don't want it to fail silently, we wouldn't notice
```suggestion
          throw err
```

juliangruber (changes requested):

Issue #42, opened 2025-05-30
Map (Client, CommP) to the URL where to retrieve from


bajtos:
1. Find which ProofSets contain the requested CommP
2. Pick one of the ProofSets to use
3. Find the owner address (SP wallet address) of the ProofSet
4. Map the SP wallet address to retrieval URL

bajtos:
Since we don't have access to the data linking PDP deals (ProofSets) to clients, I propose to implement the following interim solution:

- Find _all_ roots matching the requested CID.
- Join those roots with their ProofSets.
- Now we have a list of records `(root_cid, root_id, set_id, owner)`
- Filter the records and keep only the entries where the `owner` is one of the SPs participating in the hackathon.
- If there is more than one entry left, we need an algorithm to decide where to fetch from. Two ideas to consider (feel free to suggest something else!):
   - We can choose at random. This will distribute the load across all SPs. However, it means requests may become inconsistent: e.g. the first request is handled by SP1 which fails the retrieval, and the second request gets routed to SP2 which serves the file.
   - We can sort providers using their owner address and always pick the first one. This is predictable and consistent, but may put disproportionately large load on a single SP.

@juliangruber @NikolasHaimerl thoughts?

bajtos:
> Do we expect this to be a solution only for June 6.?

Yes.

The goal is to reduce the risk of not shipping anything by June 6 in case the Pandora contract & Synapse SDK are shipped too late and we don't manage to integrate in time.

My proposal above describes a solution to make FilCDN usable for hackathon participants even if Pandora & Synapse are not ready.

I also see it as the first step of on a two-step incremental journey:
1. First, make sure we can retrieve any PDP deal as long as it's stored with one of the SPs we support
2. Second, tune the algorithm selecting the SP to take into account the client. (I.e. tweak the filter in step 4 to look at ProofSets of the client only.)

PR #74, opened 2025-06-05
feat: url mapping

NikolasHaimerl:
### 🎯 Purpose
This PR introduces dynamic routing of file retrieval requests based on the **owner address of a root CID**, replacing the previous hardcoded service provider (SP) URL. It ensures the system can route to different storage providers depending on ownership metadata stored in the database.

---

#### 🧠 Core Logic
- Added a new `OWNER_TO_URL` mapping in `retriever.js`, associating known Filecoin owner addresses with their corresponding storage provider URLs.
- Removed the hardcoded `BASE_URL` and replaced it with a dynamic lookup based on the result of `getOwnerByRootCid(env, rootCid)`.

#### 📦 Retrieval Flow Updates
- Looked up `ownerAddress` by querying `indexer_roots` and `indexer_proof_sets`.
- Routed file retrievals to the appropriate provider URL based on the resolved owner.
- Returned a `404` if the owner is unknown or not configured in the map.

#### 📐 Database Changes
- Added a new SQL migration `0005_add_root_cid_index.sql` to index `root_cid` in the `indexer_roots` table for faster lookups.

#### 🧪 Tests
- Added new integration tests for real storage providers via known `rootCid` values.
- Extended `retriever.test.js` with end-to-end tests across multiple owner→URL scenarios.
- Added unit tests for `getOwnerByRootCid()` to cover:
  - successful lookups
  - unknown `rootCid`
  - missing owner mappings

---

### 📌 Notes
- All supported PDP owner addresses for the M1 milestone are now contained in the `OWNER_TO_URL` object.
- Replaced `pieceCid` with `rootCid` in request parsing and throughout the code.

### DIFF
--- a/retriever/lib/store.js
+++ b/retriever/lib/store.js
@@ -98,7 +98,7 @@ export async function getOwnerByRootCid(env, rootCid) {
 
   if (!result) {
     return {
-      error: `Root_cid '${rootCid}' does not exist or may not be indexed yet.`,
+      error: `Root_cid '${rootCid}' does not exist or may not have been indexed yet.`,
     }
   }
 


retriever/lib/retrieval.js
```diff
@@ -13,7 +13,7 @@
  *   - The response from the fetch request, the cache miss and the content length.
  */
 export async function retrieveFile(baseUrl, pieceCid, cacheTtl = 86400) {
-  const url = `https://${baseUrl}/piece/${pieceCid}`
+  const url = `${baseUrl}/piece/${pieceCid}`
```
NikolasHaimerl, line 16: Some SPs use http, other use https

retriever/lib/store.js
```diff
@@ -67,3 +69,61 @@ export async function logRetrievalResult(env, params) {
...
+  const { owner } = ownerResult
+  assert.ok(
+    owner,
+    `Set_id '${setId}' to owner address exists but owner address is null.`,
+  )
+  assert.ok(
+    typeof owner === 'string',
+    `Set_id '${setId}' to owner address exists but owner address is not a string.`,
+  )
+
```
juliangruber, line 127:
```suggestion
```

These are impossible given the DB schema

retriever/lib/store.js
```diff
@@ -67,3 +69,61 @@ export async function logRetrievalResult(env, params) {
...
+ *   ownerAddress?: string
+ *   error?: string
+ * }>}
+ */
+export async function getOwnerByRootCid(env, rootCid) {
+  const findRootQuery = `
+    SELECT set_id FROM indexer_roots
+    WHERE root_cid = ?
+    LIMIT 1;
+  `
```
juliangruber, line 89: Here you're performing two consecutive queries, which will be slow. Could you run just one query instead, which performs the join on the sql level?

retriever/lib/store.js
```diff
@@ -67,3 +69,61 @@ export async function logRetrievalResult(env, params) {
...
+  assert.ok(
+    owner,
+    `Set_id '${setId}' to owner address exists but owner address is null.`,
+  )
+  assert.ok(
+    typeof owner === 'string',
+    `Set_id '${setId}' to owner address exists but owner address is not a string.`,
+  )
+
+  return { ownerAddress: owner }
```
juliangruber, line 128:
```suggestion
  return owner
```

retriever/test/retriever.test.js
```diff
@@ -39,6 +39,67 @@ describe('retriever.fetch', () => {
...
+    '0xCb9e86945cA31E6C3120725BF0385CBAD684040c': {
+      url: 'https://caliberation-pdp.infrafolio.com',
+      rootCid:
+        'baga6ea4seaqntcagzjqzor3qxjba2mybegc6d2jxiewxinkd72ecll6xqicqcfa',
+      proofSetId: 234,
+    },
+  }
+
+  beforeAll(async () => {
+    // Clear existing test data (optional)
```
juliangruber, line 71: Why is this optional? And if it's optional, can we delete it?

retriever/bin/retriever.js
```diff
@@ -47,15 +58,30 @@ export default {
 
     // Timestamp to measure file retrieval performance (from cache and from SP)
     const fetchStartedAt = performance.now()
-
+    const { ownerAddress, error: ownerLookupError } = await getOwnerByRootCid(
```
juliangruber, line 61: As requested below, please change to something JS native like this

```
let owner
try {
  owner = await getOwnerByRootCid() 
} catch (err) {
  // ...
}
```

retriever/bin/retriever.js
```diff
@@ -47,15 +58,30 @@ export default {
...
+      env,
+      rootCid,
+    )
+    if (ownerLookupError) {
+      console.error(ownerLookupError)
+      return new Response(ownerLookupError, { status: 404 })
+    }
+    assert.ok(ownerAddress, 'Owner address must be defined')
```
juliangruber, line 69:
```suggestion
```

It's the responsibility of the function to ensure that

retriever/lib/store.js
```diff
@@ -67,3 +69,61 @@ export async function logRetrievalResult(env, params) {
...
+    return {
+      error: `Root_cid '${rootCid}' does not exist or may not be indexed yet.`,
+    }
+  }
+  const { set_id: setId } = rootResult
+  assert.ok(setId, `Root_cid '${rootCid}' exists but has no set_id associated.`)
+  assert.ok(
+    typeof setId === 'string',
+    `Root_cid '${rootCid}' to set_id exists but set_id is not a string.`,
+  )
```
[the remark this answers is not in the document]
NikolasHaimerl: Yes, but I needed them for typescript checks.

retriever/lib/store.js
```diff
@@ -67,3 +69,61 @@ export async function logRetrievalResult(env, params) {
...
+/**
+ * Get the owner of a given root_cid. Throws an error if the root_cid is not
+ * found or mapping is incomplete. Returns default owner if mapping is absent.
+ *
+ * @param {Env} env - Worker environment (contains D1 binding).
+ * @param {string} rootCid - The root CID to look up.
+ * @returns {Promise<{
+ *   ownerAddress?: string
+ *   error?: string
+ * }>}
```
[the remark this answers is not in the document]
NikolasHaimerl: Miro followed a different approach here: https://github.com/filcdn/worker/blob/main/retriever/lib/request.js#L30
which is what I copied.

retriever/test/retriever.test.js
NikolasHaimerl: I deleted this comment

retriever/bin/retriever.js
NikolasHaimerl: It is needed for typescript checks to pass.

retriever/lib/store.js
NikolasHaimerl: Addressed in ede0167

retriever/lib/store.js
bajtos: Can we use something like an OUTER JOIN that will set columns from `indexer_proof_sets ` to null for rows that don't have a proofset linked to the root?

I think we can get afford the extra latency of two SQL queries for the M1 milestone, how about creating a follow-up tech-debt issue to improve this later, after this initial PR is shipped?

retriever/bin/retriever.js
```diff
@@ -47,15 +57,34 @@ export default {
...
+      console.error(ownerLookupError)
+      return new Response(ownerLookupError, { status: 404 })
+    }
 
+    if (
+      !ownerAddress ||
+      !Object.prototype.hasOwnProperty.call(OWNER_TO_URL, ownerAddress)
+    ) {
+      return new Response(
+        `No PDP URL configured for owner address: ${ownerAddress}`,
```
bajtos, line 75: I think this message could be confusing to people retrieving the content. I propose to rephrase it as follows:

```suggestion
        `Unsupported Storage Provider (PDP ProofSet Owner): ${ownerAddress}`,
```

retriever/lib/store.js
```diff
@@ -67,3 +67,49 @@ export async function logRetrievalResult(env, params) {
...
+      error: `Root_cid '${rootCid}' does not exist or may not be indexed yet.`,
+    }
+  }
+  const { set_id: setId } = /** @type {{ owner: string }} */ rootResult
+
+  const findOwnerQuery = `
+    SELECT owner FROM indexer_proof_sets
+    WHERE set_id = ?
+    LIMIT 1;
+  `
```
bajtos, line 102: ❗ 
I am afraid this logic is not correctly implementing the spec in https://github.com/filcdn/worker/issues/42#issuecomment-2940308442

Scenario when this fails:
- Random person trying out PDP uploads file with CID `baga1` to an SP not participating in the hackhaton (an SP we don't have the retrieval URL for).
- Hackathon participant uploads the same file (with CID `baga1`) via Synapse SDK to a supported SP
- Our query pick the `indexer_root` entry from the random person
- Then we map that root to an owner that we don't support
- And the retrieval fails.

It's essential to build the list of all SPs (owners) storing the requested CID.

bajtos (changes requested):
Looks good at high-level, most of my comments below are minor (not blocking the merge). However, I don't think the algorithm looking up the owner is correct - see the last comment below.

retriever/lib/store.js
NikolasHaimerl: That sounds good, keep in mind that both queries have an index so they should be quite fast. I am not sure whether a join is faster than two indexed queries, depends on how many values there are to be joined.

retriever/lib/store.js
NikolasHaimerl: Does it make sense to even store owner addresses that are not our supported SPs? If we filter them out at insertion we can make the queries much faster because we have fewer entries and we would need to reset the database after the hackathon anyway because we are only supporting testnet right now. @juliangruber WDYT?

retriever/bin/retriever.js
```diff
@@ -47,15 +57,34 @@ export default {
 
     // Timestamp to measure file retrieval performance (from cache and from SP)
     const fetchStartedAt = performance.now()
+    const { ownerAddress, error: ownerLookupError } =
+      /** @type {{ ownerAddress: string; error: string }} */ await getOwnerByRootCid(
```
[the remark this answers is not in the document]
NikolasHaimerl: Fixed in a6c80ec

retriever/bin/retriever.js
NikolasHaimerl: Opened an issue: https://github.com/filcdn/worker/issues/80

retriever/test/retriever.test.js
```diff
@@ -39,6 +39,66 @@ describe('retriever.fetch', () => {
     },
   }
 
+  const REAL_TEST_DATAPOINTS = {
```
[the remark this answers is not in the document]
NikolasHaimerl: Real means that the datapaoints are real datapoints that are stored on SPs.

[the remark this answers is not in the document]
juliangruber: got it, maybe rename to `REAL_DATAPOINTS` then? right now they are real _test_ datapoints

retriever/lib/constants.js
```diff
@@ -0,0 +1,35 @@
...
+ * @type {Record<
+ *   string,
+ *   { url: string; rootCid: string; proofSetId: number }
+ * >}
+ */
+export const OWNER_TO_RETRIEVAL_URL_MAPPING = {
+  '0x12191de399B9B3FfEB562861f9eD62ea8da18AE5': {
+    url: 'https://techx-pdp.filecoin.no',
+    rootCid: 'baga6ea4seaqmqjamoiors6rjncefkohlqd2yw7k5ockt2u5fkr6d6rcwpfp5ejq',
+    proofSetId: 239,
```
juliangruber, line 12: why do these include root cid and proof set id? Every owner has more than one

NikolasHaimerl: Yes, I consolidated this data so it can be used both for the mapping used in retriever.js and in the testing, otherwise, I would have had to hardcode the addresses twice, which is what I tried to resolve from the earlier comment.

NikolasHaimerl: I also mentioned this in the description of the constant

retriever/test/retriever.test.js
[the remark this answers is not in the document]
NikolasHaimerl: I removed these datapoints when creating the consilidated constant in `constants.js` and marked them with `sample` to indicate that they are sample datapoints of the SP

retriever/lib/store.js
```diff
@@ -67,3 +69,61 @@ export async function logRetrievalResult(env, params) {
...
+    WHERE root_cid = ?
+    LIMIT 1;
+  `
+
+  const rootResult = await env.DB.prepare(findRootQuery).bind(rootCid).first()
+
+  if (!rootResult) {
+    return {
+      error: `Root_cid '${rootCid}' does not exist or may not be indexed yet.`,
+    }
```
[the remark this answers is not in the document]
NikolasHaimerl: Addressed in 41b383f

retriever/lib/retrieval.js
pyropy, line 16:
```suggestion
  const url = `${baseUrl}/piece/${rootCid}`
```

Maybe let's rename `pieceCid` to `rootCid` here to keep it consistent with rest of the codebase.

juliangruber (approved):
Looks great to me! 👏

PR #83, opened 2025-06-09
fix: add owner case insensitiveness

NikolasHaimerl:
#### ✨ Summary

This PR ensures that owner addresses (used to identify PDP storage providers) are treated in a **case-insensitive** manner throughout the indexing and retrieval flow. This fixes potential mismatches between the on-chain mixed-case Ethereum addresses and the normalised lowercase keys in our internal mapping.

---

#### ✅ Changes Made

1. **Indexer (`indexer.js`)**

   * Owner addresses are now stored in lowercase using `.toLowerCase()` when received via `/proof-set-created`.

2. **Constants (`constants.js`)**

   * All keys in `OWNER_TO_RETRIEVAL_URL_MAPPING` are converted to lowercase for consistent lookup.

3. **Store Logic (`store.js`)**

   * Normalized owner addresses (to lowercase) when checking approved owners.
   * Ensured returned `ownerAddress` is always in lowercase format.

4. **Test Coverage**

   * ✅ Unit test in `store.test.js` verifying `getOwnerByRootCid` handles mixed-case owners correctly.
   * ✅ Unit test in `indexer.test.js` verifying indexer handles mixed-case owners correctly.
   * ✅ Integration test in `retriever.test.js` confirming retrieval succeeds with mixed-case addresses stored in the DB.
   * Updated existing tests to reflect the normalized return format.

---

#### 🧪 Why This Matters

Ethereum addresses are **not case-sensitive**, but our previous implementation assumed strict equality. This led to failed matches if a mixed-case address was inserted or queried.

By normalizing all owner address inputs and lookups to lowercase:

* 🔒 We avoid mismatched lookups
* 🚀 Ensure consistent behavior across environments
* ✅ Conform to best practices for address handling

---

#### ✅ Test Plan

* Ran `store.test.js` and `retriever.test.js`
* Confirmed all tests pass
* Added new test cases for case-insensitive behavior


juliangruber (approved):
looks great to me, let's ship

retriever/test/retriever.test.js
```diff
@@ -316,6 +316,61 @@ describe('retriever.fetch', () => {
...
+      'INSERT INTO indexer_proof_sets (set_id, owner) VALUES (?, ?)',
+    )
+      .bind(proofSetId, mixedCaseOwner)
+      .run()
+
+    await env.DB.prepare(
+      'INSERT INTO indexer_roots (root_id, set_id, root_cid) VALUES (?, ?, ?)',
+    )
+      .bind('root-case-test', proofSetId, rootCid)
+      .run()
```
pyropy, line 338: We can extract this into a function like `withProofSet` similar to the `withRequest` helper function that we use.

retriever/lib/store.js
```diff
@@ -103,18 +105,18 @@ export async function getOwnerByRootCid(env, rootCid) {
   }
 
   const { set_id: setId, owner } = result
-
-  if (owner === null) {
+  const ownerCaseInsensitive = owner?.toLowerCase()
```
pyropy, line 108: We could also convert owner address to lowercase on the query level using [lower()](https://www.w3resource.com/sqlite/core-functions-lower.php) function.

pyropy (commented):
Code change looks good, I'd move owner address normalization to query level before we merge this.

pyropy:
> Code change looks good, I'd move owner address normalization to query level before we merge this.

Aslo we should probably normalise the data on insertion rather then on the query level but either one should work.

retriever/test/retriever.test.js
NikolasHaimerl: I think this is part of a different PR if we want to do this. The reason why this line shows up is because I added a new test with a similar pattern as the other tests.

retriever/lib/store.js
NikolasHaimerl: Done in deb1a48

retriever/test/retriever.test.js
pyropy: @NikolasHaimerl Would you mind creating a follow-up issue for that and mark it as tech-debt?

retriever/test/store.test.js
```diff
@@ -98,7 +98,7 @@ describe('getOwnerByRootCid', () => {
     ])
 
     const result = await getOwnerByRootCid(env, cid)
-
+    console.log('Result:', result)
```
pyropy, line 101: I assume this has been pushed on accident.

NikolasHaimerl: yes, I'll remove it.

PR #87, opened 2025-06-11
fix owner lookup

NikolasHaimerl:
### 📄 Description

This PR fixes a logic flaw in `getOwnerByRootCid` where the function previously returned the **first matching owner** from SQL for a given `root_cid`, without checking if the owner was in the **approved list**.

#### 🐞 Before:

* If two different storage providers (SPs) shared the same `root_cid` (common in re-announcements or duplicated proofs),
* And only one of them was approved,
* The function could mistakenly return the **unapproved** one — simply because it was returned first by SQL.

#### ✅ After:

* The function now:

  * Retrieves **all matching entries** for the `root_cid`
  * Filters out null/undefined `owner`s
  * Selects the **first owner in the `approvedOwners` list**, ensuring valid attribution
  * Returns meaningful errors if:

    * No owners exist for the CID
    * No approved owners exist

---

### 🧪 Tests Added

* ✔ Validates correct owner returned when multiple SPs share a `root_cid` but only one is approved
* ✔ Adds edge cases for:

  * No owner present
  * Only unapproved owner
  * Two owners exist of which only one is approved

---

### DIFF
--- a/retriever/lib/store.js
+++ b/retriever/lib/store.js
@@ -122,7 +122,7 @@ export async function getOwnerByRootCid(env, rootCid) {
   const { set_id: setId, owner } = approved
 
   console.log(
-    `Retrieved set_id '${setId}' and owner '${owner}' for root_cid '${rootCid}'`,
+    `Looked up set_id '${setId}' and owner '${owner}' for root_cid '${rootCid}'`,
   )
 
   return { ownerAddress: owner }


retriever/lib/store.js
```diff
@@ -92,33 +92,43 @@ export async function getOwnerByRootCid(env, rootCid) {
...
     }
   }
+  const { set_id: setId, owner } = approved
+
+  console.log(
+    `Retrieved set_id '${setId}' and owner '${owner}' for root_cid '${rootCid}'`,
+  )
 
-  if (!approvedOwners.includes(owner)) {
+  if (!approvedOwners.includes(String(owner))) {
```
bajtos, line 127: Why is this case necessary? What other value type than `string` can the `owner` property contain?

Can we fix the types earlier, so that TypeScript understands here and below that `owner` is a `string`?

retriever/lib/store.js
```diff
@@ -92,33 +92,43 @@ export async function getOwnerByRootCid(env, rootCid) {
...
     return {
       error: `Set_id '${setId}' is associated with owner '${owner}', which is none of the currently supported SPs.`,
     }
   }
 
-  return { ownerAddress: owner }
+  return { ownerAddress: String(owner) }
```
bajtos, line 133: Ditto.

bajtos (approved):
Looks good to me.

There is one place where we should improve our type annotations, but that can be fixed after this pull request is landed, if you prefer.

retriever/lib/store.js
```diff
@@ -92,33 +92,43 @@ export async function getOwnerByRootCid(env, rootCid) {
    LEFT OUTER JOIN indexer_proof_sets ips
      ON ir.set_id = ips.set_id
    WHERE ir.root_cid = ?
-   LIMIT 1;
  `
 
-  /** @type {{ set_id: string; owner: string | null } | null} */
-  const result = await env.DB.prepare(query).bind(rootCid).first()
+  const result = await (await env.DB.prepare(query).bind(rootCid).all()).results
```
juliangruber, line 97:
```suggestion
  const results = await (await env.DB.prepare(query).bind(rootCid).all()).results
```

Should be called `results`, as it is an array now. Btw, there are too many `awaits` here, it should be like this then:

```suggestion
  const results = (await env.DB.prepare(query).bind(rootCid).all()).results
```

And simplifier with more modern syntax:

```suggestion
  const { results } = await env.DB.prepare(query).bind(rootCid).all()
```

retriever/lib/store.js
```diff
@@ -92,33 +92,43 @@ export async function getOwnerByRootCid(env, rootCid) {
...
+  const withOwner = result.filter((row) => row && row.owner != null)
+  if (withOwner.length === 0) {
+    return {
+      error: `Root_cid '${rootCid}' exists but has no associated owner.`,
+    }
+  }
+  const approved = withOwner.find(
+    (row) =>
+      typeof row.owner === 'string' &&
+      approvedOwners.includes(String(row.owner)),
```
juliangruber, line 114: Can we work with typescript to not need this string cast?

retriever/lib/store.js
```diff
@@ -92,33 +92,43 @@ export async function getOwnerByRootCid(env, rootCid) {
...
   )
-  if (owner === null || owner === undefined) {
+  if (!approved || approved.length === 0) {
     return {
-      error: `Set_id '${setId}' exists but has no associated owner.`,
+      error: `Root_cid '${rootCid}' exists but has no associated owner from the approved list.`,
```
juliangruber, line 118:
```suggestion
      error: `Root_cid '${rootCid}' exists but has no approved owner.`,
```

juliangruber (changes requested):

retriever/lib/store.js
bajtos: FWIW, we are discussing this in https://github.com/filcdn/worker/pull/87/files#r2139928321

retriever/lib/store.js
bajtos: The problem is that `result` has the type `Record<string, unknown>[]` - i.e. an array of objects (rows) where each row has string-named columns with unknown-typed values.

The solution is to tell TypeScript about the data type of the rows returned by the SQL query.

If we were using TypeScript, you could write `DB...all<{ owner: string; set_id: string }>()`. Unfortunately, JSDoc comments don't support specifying parameterised return values.

Here is the code that works for me:

```js
 const result = /** @type {{ owner: string; set_id: string }[]} */ (
    /** @type {any[]} */ (
      (await env.DB.prepare(query).bind(rootCid).all()).results
    )
  )
```

Once you made this change, typescript will flag an error on this line:

```js
  if (!approved || approved.length === 0) {
```

Reason: `Array.prototype.find` returns either one array item or `undefined`. I think your condition assumes `approved` is an array. (But the code later treats `approved` as a single object.)

I am writing this in length to show why it's better to think more deeply about TypeScript errors and not take the lazy approach.

bajtos (changes requested):

retriever/lib/store.js
NikolasHaimerl: Addressed in 567a17c

retriever/lib/store.js
NikolasHaimerl: Addressed in 567a17c

NikolasHaimerl:
https://github.com/filcdn/worker/pull/87#discussion_r2139993472

@bajtos thanks for providing the correct typing, I did try the first part but missed the `any[]` typing, not sure why there is a necessity for casting the type twice and use `any` as an intermediate. I usually try to avoid `any`.

bajtos (changes requested):
Almost there!

retriever/lib/store.js
```diff
@@ -94,41 +94,45 @@ export async function getOwnerByRootCid(env, rootCid) {
...
     return {
-      error: `Root_cid '${rootCid}' exists but has no associated owner from the approved list.`,
+      error: `Root_cid '${rootCid}' exists but has no approved owner`,
     }
   }
+
+  if (!approved || typeof approved.owner !== 'string') {
+    // Should never be called
+    return { error: `Record has unexpected format ${JSON.stringify(approved)}` }
+  }
```
juliangruber: Is this required for TS? Can we remove that dead block?

bajtos: I think it may have been needed while `approved` had type `Record<string, unknown>`. That that it's typed, I think this shouldn't be needed anymore.

bajtos: (I am also surprised `tsc` did not flag this.)

retriever/lib/store.js
```diff
@@ -94,41 +94,42 @@ export async function getOwnerByRootCid(env, rootCid) {
...
   const { set_id: setId, owner } = approved
 
   console.log(
     `Retrieved set_id '${setId}' and owner '${owner}' for root_cid '${rootCid}'`,
   )
 
-  if (!approvedOwners.includes(String(owner))) {
+  if (!approvedOwners.includes(owner)) {
     return {
       error: `Set_id '${setId}' is associated with owner '${owner}', which is none of the currently supported SPs.`,
```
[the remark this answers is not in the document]
NikolasHaimerl: Removed in 7ba967f

retriever/lib/store.js
```diff
@@ -92,33 +92,38 @@ export async function getOwnerByRootCid(env, rootCid) {
...
     return {
-      error: `Set_id '${setId}' is associated with owner '${owner}', which is none of the currently supported SPs.`,
+      error: `Root_cid '${rootCid}' exists but has no approved owner`,
     }
   }
 
+  const { set_id: setId, owner } = approved
+
+  console.log(
+    `Retrieved set_id '${setId}' and owner '${owner}' for root_cid '${rootCid}'`,
```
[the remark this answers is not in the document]
bajtos: Great point!

bajtos:
@NikolasHaimerl I am going to land this pull request myself to silence the false alarm notifications in #alerts channel.

bajtos:
Ah, I see you enabled auto-merge, please ignore my previous comment.

PR #90, opened 2025-06-12
retriever: Refactor error handling using `http-assert`-like fn. Closes #80

### LINKED ISSUE #80: Change error pattern in retriever
### Background

In several places in the codebase, async function calls are handled using a destructuring pattern that includes the error as part of the response object. For example:

```js
const { ownerAddress, error: ownerLookupError } = await getOwnerByRootCid()
````

While this pattern may work functionally, it is non-standard in JavaScript and can reduce readability for those unfamiliar with the convention.

### Proposal

Refactor these occurrences to use JavaScript's native `try/catch` pattern, which is clearer and more idiomatic. For example, the above code should be rewritten as:

```js
let owner
try {
  owner = await getOwnerByRootCid()
} catch (err) {
  // handle error
}
```

### Benefits

* Improves code readability and maintainability
* Aligns with widely accepted JavaScript conventions
* Reduces potential confusion for new contributors

### Scope

* Identify and refactor all instances in the codebase where error handling is done via destructuring async results
* Replace them with `try/catch` blocks where appropriate
* Ensure existing behavior is preserved
* Update or add tests if necessary

### Notes

This was discussed during the review of a recent PR, but was deemed out of scope for that change. This issue tracks the proposed refactor as a follow-up.

### References

* [[Related PR discussion](https://github.com/filcdn/worker/pull/74#discussion_r2128702211)]

juliangruber:
Closes #80

Using language native errors instead of custom `{ error }` properties, which are easier to miss

### DIFF
--- a/retriever/bin/retriever.js
+++ b/retriever/bin/retriever.js
@@ -6,6 +6,7 @@ import {
   measureStreamedEgress,
 } from '../lib/retrieval.js'
 import { getOwnerByRootCid, logRetrievalResult } from '../lib/store.js'
+import { httpAssert } from '../lib/http-assert.js'
 
 export default {
   /**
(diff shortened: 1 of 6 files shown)


retriever/bin/retriever.js
```diff
@@ -17,56 +18,63 @@ export default {
...
-    const {
-      clientWalletAddress,
-      rootCid,
-      error: parsingError,
-    } = parseRequest(request, env)
-    if (parsingError) {
-      console.error(parsingError)
-      return new Response(parsingError, { status: 400 })
-    }
+    const { clientWalletAddress, rootCid } = parseRequest(request, env)
```
juliangruber, line 56: @bajtos through forwarding errors (from `_fetch` to `fetch`), we don't need to write a `try/catch` for every call. This is the main lesson from using something like `http-assert` and then having a global error handling (here: `fetch`)

retriever/lib/http-assert.js
juliangruber, line 1: I had to inline this library, as cloudflare workers threw syntax errors when I imported `http-assert` from npm

retriever/lib/request.js
```diff
@@ -7,28 +9,27 @@
...
-    return { error: 'Missing required path element: `/{CID}`' }
-  }
 
-  if (!rootCid.startsWith('baga')) {
-    return { error: `Invalid CID: ${rootCid}. It is not a valid CommP root.` }
-  }
+  httpAssert(rootCid, 400, 'Missing required path element: `/{CID}`')
+  httpAssert(
+    rootCid.startsWith('baga'),
+    400,
```
bajtos, line 30: Same here. _(Feel free to ignore.)_

```suggestion
    404,
```

bajtos (changes requested):
Great improvement!

retriever/bin/retriever.js
```diff
@@ -17,56 +18,63 @@ export default {
...
+      return await this._fetch(request, env, ctx, retrieveFile)
+    } catch (error) {
+      console.error(error)
+      const message =
+        typeof error === 'object' &&
+        error !== null &&
+        'message' in error &&
+        typeof error.message === 'string'
+          ? error.message
+          : String(error)
```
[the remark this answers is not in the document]
juliangruber: Good catches! Addressed in 0418a20 and 55c68d5

retriever/bin/retriever.js
```diff
@@ -17,56 +18,61 @@ export default {
...
+    } catch (error) {
+      const errHasStatus =
+        typeof error === 'object' &&
+        error !== null &&
+        'status' in error &&
+        typeof error.status === 'number'
+      const message =
+        errHasStatus && 'message' in error && typeof error.message === 'string'
+          ? error.message
+          : 'Internal Server Error'
```
NikolasHaimerl, line 32: Is Internal Server Error descriptive enough?

juliangruber: what else would you suggest? This is an unexpected error

retriever/bin/retriever.js
```diff
@@ -21,21 +21,19 @@ export default {
...
-      const status =
+      const errHasStatus =
         typeof error === 'object' &&
         error !== null &&
         'status' in error &&
         typeof error.status === 'number'
-          ? error.status
-          : 500
+      const message =
+        errHasStatus && 'message' in error && typeof error.message === 'string'
```
bajtos, line 30: Internal server errors (5xx) can contain sensitive information (e.g. "connection refused" often includes the hostname of the remote server); therefore, we shouldn't be exposing the error message to the clients.

On the other hand, I can see an argument that if the caller specified `error.status = 5xx` then they probably specified also a custom error message that should be safe to send to the clients. (But this will be difficult to enforce in practice.)

WDYT?

```suggestion
        errHasStatus && error.status < 500 && 'message' in error && typeof error.message === 'string'
```

bajtos (approved):

retriever/bin/retriever.js
```diff
@@ -17,56 +18,61 @@ export default {
    * @returns
    */
   async fetch(request, env, ctx, { retrieveFile = defaultRetrieveFile } = {}) {
```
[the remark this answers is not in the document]
NikolasHaimerl: I see let's move this discussion then to the framework selection.

bajtos:
@juliangruber I fixed the TypeScript error in [ad1d6a6](https://github.com/filcdn/worker/pull/90/commits/ad1d6a632d46608dce3a9b9a309f59909e21f051). I also decided to extract the error handler code into another method.

retriever/bin/retriever.js
bajtos: The "Internal Server Error" branch is triggered when something unexpected happens, typically due to a programmer's error - the code fails to handle an edge case that it was supposed to handle. This may be triggered by client-side behaviour, but it's not something we want to tell clients about.

PR #98, opened 2025-06-18
chore: use built in function for cloudflare D1 databases

### LINKED ISSUE #57: Use Cloudflare solution for applying DB migrations in tests
Docs:
https://developers.cloudflare.com/workers/testing/vitest-integration/test-apis/#d1

Example:
https://github.com/cloudflare/workers-sdk/tree/main/fixtures/vitest-pool-workers-examples/d1

NikolasHaimerl:
Closes  #57


indexer/vitest.config.js
```diff
@@ -1,12 +1,31 @@
-import { defineWorkersConfig } from '@cloudflare/vitest-pool-workers/config'
+import path from 'node:path'
+import {
+  defineWorkersProject,
+  // @ts-ignore
```
NikolasHaimerl, line 4: lint complains about the import not existing. The example from cloudflare [imports](https://github.com/cloudflare/workers-sdk/blob/main/fixtures/vitest-pool-workers-examples/d1/vitest.config.ts) it this way. 

I believe it might be due to the fact that we are using javascript instead of typescript here. The cloudflare example is also using typescript.

bajtos:
Great job, @NikolasHaimerl! I'll take a closer look tomorrow or on Monday.

indexer/vitest.config.js
bajtos: Import map for `@cloudflare/vitest-pool-workers/config`:
https://github.com/cloudflare/workers-sdk/blob/17d23d8e5fd54737d1c4b9cb487fd6e85cddc9c8/packages/vitest-pool-workers/package.json#L34-L38

Entry point:
https://github.com/cloudflare/workers-sdk/blob/main/packages/vitest-pool-workers/src/config/index.ts

`readD1Migrations` is defined (and exported) here:
https://github.com/cloudflare/workers-sdk/blob/17d23d8e5fd54737d1c4b9cb487fd6e85cddc9c8/packages/vitest-pool-workers/src/config/d1.ts#L9

here it is re-exported by the index:

https://github.com/cloudflare/workers-sdk/blob/17d23d8e5fd54737d1c4b9cb487fd6e85cddc9c8/packages/vitest-pool-workers/src/config/index.ts#L234

bajtos: Let me merge the current `main` branch to bring in [bump @cloudflare/vitest-pool-workers from 0.8.39 to 0.8.41](https://github.com/filcdn/worker/pull/100)

NikolasHaimerl: Yes, I had the same issue, it also runs just fine when running `npm test` without the `ts-ignore` but the lint always fails.

NikolasHaimerl: Does that mean it has to do with the fact that we are using JS instead of TS?

bajtos (approved):
The new test setup works great on my machine. 👏🏻 Thank you for the cleanup, @NikolasHaimerl!

Issue #41, opened 2025-05-30
Validate that the client has a PDP deal storing the requested CommP and `withCDN` flag is enabled


pyropy:
Lock down retrievals - clients can retrieve only Roots stored in client’s ProofSets with CDN enabled. This also prevents HTTP clients from retrieving the piece from an SP that the client doesn’t have a CDN deal with

- [x] Implement subgraph+webhook to ingest information about active deals `(proofSetId, clientWalletAddress, withCDN)` to our Cloudflare database.
- [x] Validate that requested CommP for client is part of a PDP deal
- [x] Validate that the PDP deal storing the requested CommP has `withCDN` flag enabled. Return [402 Payment Required](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status/402) when `withCDN` is not enabled.
- [x] Announce the change in hackathon Discord and other public channels.

Pre-requisites:
- https://github.com/filcdn/roadmap/issues/6

PR #101, opened 2025-06-19
fix: store proof_set_rails ID columns as integers

bajtos:
Before this change, our calibration database on Cloudflare contained ID values like `332.0`.

This commit converts numbers to strings on the JavaScript side and writes ID column values like `332`.

Links:
- https://github.com/filcdn/worker/issues/41

### DIFF
--- a/indexer/bin/indexer.js
+++ b/indexer/bin/indexer.js
@@ -135,7 +135,15 @@ export default {
     } else if (pathname === '/proof-set-rail-created') {
       if (
         !payload.proof_set_id ||
+        !(
+          typeof payload.proof_set_id === 'number' ||
+          typeof payload.proof_set_id === 'string'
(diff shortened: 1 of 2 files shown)


bajtos:
> I fixed existing rail records by running the following SQL query:
> 
> ```sql
> UPDATE indexer_proof_set_rails SET 
>   proof_set_id = RTRIM(proof_set_id, '.0'), 
>   rail_id = RTRIM(rail_id, '.0')
> ```

Ouch, that query corrupted rows for IDs ending with 0.

```
SELECT RTRIM('340.0', '.0')
> 34
```

bajtos:
This CLI seems to trigger restart from scratch:

```
goldsky pipeline restart pdp-pandora-proof-set-rail-created --from-snapshot none
```

PR #110, opened 2025-06-24
feat: add provider registered webhook

NikolasHaimerl:
Adds an endpoint for `provider-registered`. 
This endpoint will be called by the Goldsky Pipeline. It will add the event for registering providers. 
See: https://github.com/FilOzone/filecoin-services/blob/main/service_contracts/src/PandoraService.sol#L144C63-L144C69

Related to https://github.com/filcdn/roadmap/issues/13


indexer/test/indexer.test.js
```diff
@@ -436,4 +436,88 @@ describe('retriever.indexer', () => {
...
+      const providerUrl = 'https://provider.example.com'
+      const owner = '0xOwnerAddress'
+      const req = new Request('https://host/provider-registered', {
+        method: 'POST',
+        headers: {
+          [env.SECRET_HEADER_KEY]: env.SECRET_HEADER_VALUE,
+        },
+        body: JSON.stringify({
+          provider: owner,
+          pdpUrl: providerUrl,
```
juliangruber, line 462: These are confusing, let's use one term everywhere

indexer/lib/handler.js
```diff
@@ -0,0 +1,36 @@
...
+    `
+        INSERT INTO owner_urls (
+          owner,
+          url
+        )
+        VALUES (?, ?)
+        ON CONFLICT(owner) DO UPDATE SET url=excluded.url
+      `,
+  )
+    .bind(provider.toLowerCase(), pdpUrl)
```
juliangruber, line 32: The `provider.toLowerCase()` needs a test

db/migrations/0007_add_owner_to_url_table.sql
```diff
@@ -0,0 +1,6 @@
+CREATE TABLE IF NOT EXISTS owner_urls (
+  owner TEXT NOT NULL,
```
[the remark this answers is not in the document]
juliangruber: Also if we rename these fields, update the table name to match.

db/migrations/0007_add_owner_to_url_table.sql
```diff
@@ -0,0 +1,6 @@
...
+  url TEXT NOT NULL,
+  PRIMARY KEY (owner),
+  CONSTRAINT unique_owner_url UNIQUE (owner, url)
```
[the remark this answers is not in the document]
NikolasHaimerl: Yes it is. I guess for now it is not needed.

indexer/bin/indexer.js
```diff
@@ -175,6 +176,9 @@ export default {
         )
         .run()
       return new Response('OK', { status: 200 })
+    } else if (pathname === '/provider-registered') {
+      const { provider, pdpUrl } = payload
```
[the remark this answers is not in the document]
NikolasHaimerl: The event emitted from the pandora [contract](https://github.com/FilOzone/filecoin-services/blob/main/service_contracts/src/PandoraService.sol#L144) is `pdpUrl` which is why I assumed that GoldSky also uses `pdpUrl`. Do you know whether GoldSky changes the field's names?

db/migrations/0007_add_owner_to_url_table.sql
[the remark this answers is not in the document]
NikolasHaimerl: I was debating the same thing. We are using owner inside of filCDN and all tables, which is why I went with that. 
I don't have a strong opinion on this. I thought I'd go for consistency within filCDN.

db/migrations/0007_add_owner_to_url_table.sql
```diff
@@ -0,0 +1,6 @@
+CREATE TABLE IF NOT EXISTS owner_urls (
+  owner TEXT NOT NULL,
+  url TEXT NOT NULL,
```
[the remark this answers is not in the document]
NikolasHaimerl: Addressed in 2594cd6

indexer/lib/handler.js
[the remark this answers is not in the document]
NikolasHaimerl: Addressed in 2594cd6

db/migrations/0007_add_owner_to_url_table.sql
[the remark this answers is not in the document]
NikolasHaimerl: Addressed in 2594cd6

indexer/lib/handler.js
NikolasHaimerl: Added in 2e60df3

indexer/bin/indexer.js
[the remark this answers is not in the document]
juliangruber: I think so, see above code, which also uses under_scores. I think they convert camel case to snake case

juliangruber (changes requested):
but please verify whether it's `pdp_url` or `pdpUrl`

indexer/lib/handle-provider-events.js
```diff
@@ -0,0 +1,36 @@
+/**
+ * Handles the /provider-registered webhook
+ *
+ * @param {Env} env
+ * @param {string} provider
+ * @param {string} pdpUrl
+ * @returns {Promise<Response>}
+ */
+export async function handleProviderRegistered(env, provider, pdpUrl) {
+  if (
```
pyropy, line 10: We also validate if `provider` is a valid eth address (at least valid format of eth address) and that `pdpUrl` is valid a http/https URL.

db/migrations/0007_add_owner_urls.sql
```diff
@@ -0,0 +1,5 @@
+CREATE TABLE IF NOT EXISTS owner_urls (
```
pyropy, line 1: I suggest that we name this table `provider_urls` or `provider_details` to match the Pandora service and indexer naming. Also I think it makes more sense then `owner_urls` as `owner` is some address that owns some (storage) provider.

NikolasHaimerl: I am not sure it does. Owner in this case is the same as owner in all other tables, which is why I went with this name. If we went with provider as the new naming than we have two names for the same thing in our database which is what I would like to avoid.

indexer/bin/indexer.js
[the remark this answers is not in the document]
NikolasHaimerl: Also when I query the event directly it is pdpUrl: 
image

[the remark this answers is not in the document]
NikolasHaimerl: Ah when I go into the schema it then changes to: 
image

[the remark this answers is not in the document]
NikolasHaimerl: I changed `pdpUrl` to `pdp_url`

indexer/lib/handle-provider-events.js
pyropy: I propose to validate if `provider` is valid eth address and `pdpUrl` is valid url.

db/migrations/0007_add_owner_urls.sql
```diff
@@ -0,0 +1,5 @@
...
+  owner TEXT NOT NULL,
```
bajtos, line 2: Should we use "owner_address" to make it clearer what's stored in this column?

db/migrations/0007_add_owner_urls.sql
```diff
@@ -0,0 +1,5 @@
...
+  pdp_url TEXT NOT NULL,
```
bajtos, line 3: IIRC, there are two URLs in the Pandora registry of SPs: pdpUrl is used for uploading new data and pieceRetrievalUrl is used for retrievals.

See https://github.com/FilOzone/filecoin-services/blob/1b99d8b39bf7f35608884b0bd7bd768810ee2372/service_contracts/src/PandoraService.sol#L119

For FilCDN, we need the latter.

Let's name the column accordingly to prevent confusion.

```suggestion
  piece_retrieval_url TEXT NOT NULL,
```

indexer/bin/indexer.js
```diff
@@ -1,3 +1,4 @@
+import { handleProviderRegistered } from '../lib/handle-provider-events.js'
```
bajtos, line 1: We use nouns, not verbs, for naming files.

```suggestion
import { handleProviderRegistered } from '../lib/provider-events-handler.js'
```

indexer/bin/indexer.js
```diff
@@ -175,6 +176,9 @@ export default {
         )
         .run()
       return new Response('OK', { status: 200 })
+    } else if (pathname === '/provider-registered') {
+      const { provider, pdp_url: pdpUrl } = payload
```
bajtos, line 180: See my other comment, we need `pieceRetrievalUrl`.

indexer/lib/handle-provider-events.js
```diff
@@ -0,0 +1,36 @@
+/**
+ * Handles the /provider-registered webhook
+ *
+ * @param {Env} env
+ * @param {string} provider
+ * @param {string} pdpUrl
```
bajtos, line 6: Ditto.

```suggestion
 * @param {string} pieceRetrievalUrl
```

indexer/lib/handle-provider-events.js
bajtos: Yeah, I think it's a great idea to add the validation here. When the validation fails, we need to alert us on Slack. 

- In this pull request, it's enough to print a structured log message via console.error.
- Then edit the description of the following issue and add another todo-item to create an alert for this place: https://github.com/filcdn/roadmap/issues/23

db/migrations/0007_add_owner_urls.sql
bajtos: Yeah, this is a bit unfortunate, because the naming inconsistency is already present in the Pandora & PDPVerifier contracts.

- Pandora registry uses the name (storage) "provider".
- PDPVerifier uses the name "owner".
- Payment rails use the name "payee".

I think it's better to follow the names used by the smart contract we are interacting with. I.e. to have a table `provider_urls` and remember to match `provider_urls.provider_address` with `indexer_proof_sets.owner` or `indexer_proof_set_rails.payee` in our SELECT queries.

db/migrations/0007_add_owner_urls.sql
bajtos: Based on the discussion about names, I would use `provider_address` or even just `address` (considering that the table already includes `provider` in the name).

bajtos (changes requested):
Great start!

indexer/bin/indexer.js
juliangruber: This is new to me (sorry if I didn't notice), I had proposed that file name.

db/migrations/0007_add_owner_urls.sql
NikolasHaimerl: adressed in e5db03f

indexer/bin/indexer.js
NikolasHaimerl: Resolved in 19acee8

indexer/lib/handle-provider-events.js
NikolasHaimerl: 8844929

indexer/test/indexer.test.js
```diff
@@ -436,4 +436,133 @@ describe('retriever.indexer', () => {
       expect(proofSetRails[0]?.rail_id).toMatch(/^\d+$/)
     })
   })
+  describe('POST /provider-registered', () => {
```
pyropy, line 439: Could you add more tests to make sure that we cover all cases: missing provider url, missing provider address, provider address not being valid eth address, provider url not being a string or valid url?

NikolasHaimerl: I have a test for missing provider_url/provider, invalid Url. 
I'll also add one for provider_url not being a string and provider address not being an eth address.

Issue #116, opened 2025-06-25
Tests failing on CI - are we accessing storage outside of a test?


bajtos:
Quoting from https://github.com/filcdn/worker/actions/runs/15869992050/job/44744172332?pr=113#step:6:798

```
Failed to pop isolated storage stack frame in test/retrieval.test.js's test "uses the default cacheTtl if not provided".
In particular, we were unable to pop D1 storage.
This usually means your Worker tried to access storage outside of a test, or some resources have not been disposed of properly.
Ensure you "await" all Promises that read or write to these services, and make sure you use the "using" keyword when passing data across JSRPC.
See https://developers.cloudflare.com/workers/testing/vitest-integration/known-issues/#isolated-storage for more details.
```

PR #117, opened 2025-06-25
fix: add abort signal to fetch method

NikolasHaimerl:
Closes https://github.com/filcdn/worker/issues/116

The issue was that the [test](https://github.com/filcdn/worker/blob/main/retriever/test/retriever.test.js#L298) for real storage providers did not dispose the pending requests causing the test suite to fail some times. 
This PR is fixing this by introducing an abort signal that can be controlled from within the test.

### DIFF
--- a/retriever/bin/retriever.js
+++ b/retriever/bin/retriever.js
@@ -40,7 +40,12 @@ export default {
    * @param {typeof defaultRetrieveFile} [options.retrieveFile]
    * @returns
    */
-  async _fetch(request, env, ctx, { retrieveFile, signal } = {}) {
+  async _fetch(
+    request,
+    env,
+    ctx,
+    { retrieveFile = defaultRetrieveFile, signal } = {},
+  ) {
     const requestTimestamp = new Date().toISOString()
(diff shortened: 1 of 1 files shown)


retriever/bin/retriever.js
```diff
@@ -30,9 +36,11 @@ export default {
    * @param {Env} env
    * @param {ExecutionContext} ctx
    * @param {typeof defaultRetrieveFile} retrieveFile
+   * @param {object} options
+   * @param {AbortSignal} [options.signal]
    * @returns
    */
-  async _fetch(request, env, ctx, retrieveFile) {
+  async _fetch(request, env, ctx, retrieveFile, { signal } = {}) {
```
bajtos, line 43: Can we use the same params as in the public-facing `fetch` method?

```suggestion
  async _fetch(request, env, ctx, {retrieveFile, signal }) {
```

bajtos (changes requested):
Nice find! 👏🏻 

The changes look great except one detail about an internal API, see my comment below.

Issue #126, opened 2025-07-01
Database Schema for BadBits and Content Status


NikolasHaimerl:
Create the database schema needed to store BadBits denylist hashes and content status.

## Description

We need to extend our database to store BadBits denylist entries and track the status of content CIDs in relation to this denylist.

## Tasks

- [ ] Create migration script for the badbits table
- [ ] Add a status column to the indexer_roots table

### Migration Script

The migration should create the following changes:

```sql
-- Add status column to indexer_roots table
ALTER TABLE indexer_roots ADD COLUMN status TEXT DEFAULT 'unchecked' CHECK (status IN ('unchecked', 'blocked', 'allowed'));

-- Create table for storing bad bits
CREATE TABLE IF NOT EXISTS badbits (
  hash TEXT PRIMARY KEY,
  hash_type TEXT NOT NULL,
  added_at DATETIME DEFAULT CURRENT_TIMESTAMP
);
```

### Status Values

The `status` column in `indexer_roots` will have three possible values:
- `unchecked`: Content has not yet been checked against the denylist
- `blocked`: Content matches an entry in the denylist and should be blocked
- `allowed`: Content has been checked and is not in the denylist

Issue #125, opened 2025-07-01
Content Moderation System with IPFS BadBits Integration


NikolasHaimerl:
This issue tracks the implementation of a content moderation system for FilCDN that integrates with the IPFS BadBits denylist to identify and block problematic content.

## Overview

We need to build a system that:
1. Regularly fetches  the BadBits denylist from IPFSand, stores and keeps them up to date
2. Computes the hash of existing CIDs to be able to compare them to this denylist. 
3. Checks existing content CIDs against this denylist
4. Update content status in our database by marking CIDs as either approved or disallowed
5. Deny CID requests where the CID is found in the denylist

This will help prevent serving content that has been reported for abuse, DMCA violations, malware, etc.

### Resources:
Documentation for blocked content: https://badbits.dwebops.pub/
Formats for double hash: https://specs.ipfs.tech/compact-denylist-format/#double-hash

PR #131, opened 2025-07-02
feat: fetch and store bad bits hashes

NikolasHaimerl:
This PR adds the database format for storing the blocked hashes from the list of blocked hashed that can be found here: https://badbits.dwebops.pub/ 
It also creates a worker which regularly fetches that list, checks in with the database whether the database is up to date with the current list and makes inserts/removes accordingly. 

In subsequent work the hashes of existing CIDs will be calculated and compared against the hashes that are retrieved from https://badbits.dwebops.pub/ . 

Closes https://github.com/filcdn/worker/issues/126
Related to 
#125 
https://github.com/filcdn/roadmap/issues/29


bad-bits/bin/scheduler.js
```diff
@@ -0,0 +1,16 @@
...
+  // Scheduled handler that runs every 5 minutes
+  async scheduled(_controller, env, _ctx) {
+    console.log('Running scheduled badbits update...')
+    try {
+      const result = await fetchAndStoreBadBits(env)
+      console.log(
+        `Updated badbits denylist: ${result.added} added, ${result.removed} removed`,
+      )
+    } catch (error) {
+      console.error('Failed to update badbits denylist:', error)
```
juliangruber, line 13:
```suggestion
      console.error('Failed to update badbits denylist:', error)
      throw err
```

Do we gain visibility if we make this throw? I assume the Cloudflare UI will then inform us of the failure

bad-bits/bin/scheduler.js
```diff
@@ -0,0 +1,16 @@
+import { fetchAndStoreBadBits } from '../lib/badbits.js'
+
+export default {
+  // Scheduled handler that runs every 5 minutes
```
juliangruber, line 4:
```suggestion
```

We want to be able to change the interval without having to update this code comment

bad-bits/lib/badbits.js
```diff
@@ -0,0 +1,41 @@
...
+
+/**
+ * @param {Env} env
+ * @param {object} options
+ * @param {typeof globalThis.fetch} [options.fetchBadBits]
+ * @returns
+ */
+export async function fetchAndStoreBadBits(
+  env,
+  { fetchBadBits } = { fetchBadBits: globalThis.fetch },
```
juliangruber, line 13: This function isn't a `fetchBadBits` function, it's a `fetch` function, accepting any URL. I propose to therefore rename it to `fetch`

juliangruber: Otherwise, change `fetchBadBits` so that you don't need to pass a URL

bad-bits/lib/badbits.js
```diff
@@ -0,0 +1,41 @@
...
+  const text = await response.text()
+  const lines = text.split('\n')
+
+  const currentBadHashes = new Set()
+
+  for (const line of lines) {
+    if (line.trim() === '' || line.startsWith('# ')) {
+      continue
+    }
+
```
juliangruber, line 31:
```suggestion
```

Having this doesn't change the outcome of the calculation, because the condition below also handles these cases

bad-bits/lib/store.js
```diff
@@ -0,0 +1,63 @@
...
+ * @param {Object} env - Environment containing database connection
+ * @param {Set<string>} currentHashes - Set of current valid hashes from
+ *   denylist
+ * @returns {Promise<{ added: number; removed: number }>} - Number of entries
+ *   added and removed
+ */
+export async function updateBadBitsDatabase(env, currentHashes) {
+  try {
+    // Get existing hashes
+    const existingHashes = new Set(await getAllBadBitHashes(env))
```
juliangruber, line 13: Instead of loading known hashes into memory, can we simply try to insert all again, and `on conflict do nothing`?

bad-bits/lib/store.js
```diff
@@ -0,0 +1,63 @@
...
+}
+
+/**
+ * Gets all bad bit hashes from the database
+ *
+ * @param {Object} env - Environment containing database connection
+ * @returns {Promise<string[]>} - Array of hash strings
+ */
+export async function getAllBadBitHashes(env) {
+  const { results } = await env.DB.prepare('SELECT hash FROM badbits').all()
```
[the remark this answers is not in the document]
juliangruber: Yes, but there's a difference between only loading the current hash list into memory, or loading the current hash list and all currently known hashes in the db too

bad-bits/lib/store.js
juliangruber: I think that will make the code simpler most of all

bad-bits/package.json
```diff
@@ -0,0 +1,16 @@
+{
+  "name": "@filcdn/bad-bits",
+  "version": "1.0.0",
+  "private": true,
+  "description": "FilCDN BadBits Worker",
+  "author": "Space Meridian <filcdn@meridian.space>",
```
juliangruber, line 6: Can you confirm this email address is set up?

juliangruber (changes requested):
good stuff!

bad-bits/lib/store.js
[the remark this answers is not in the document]
NikolasHaimerl: The hashes in the `badbits` should more or less be the same as the list of hashes retrieved from https://badbits.dwebops.pub/. This means that the amount of memory used to have both the hashes in the database and the retrieved hashes from https://badbits.dwebops.pub/ should be roughly the same. Currently the hashes from https://badbits.dwebops.pub/ are about 20MB, that means it is about 40MB in memory for both lists of hashes. I think this is reasonable, do you not think so? 

When it comes to comparing these hashes with the once from the`indexer_roots` table, I planned on checking them in batches, say 1000-10000 hashes at a time, similarly to how we check for [payload CIDs in spark deal observer](https://github.com/CheckerNetwork/spark-deal-observer/blob/main/backend/bin/deal-observer-backend.js#L132). 

Do you see an issue with proceeding like this? @juliangruber

bad-bits/bin/scheduler.js
NikolasHaimerl: I am currently working on introducing papertrail log reporting to the CDN workers. I have a draft [here](https://github.com/filcdn/worker/pull/139), which is not yet ready for review but gives you an idea of how errors will be propagated to papertrail and thus alert us on slack. 

Does this answer your question? @juliangruber

bad-bits/lib/store.js
NikolasHaimerl: This would solve the issue of newly added hashes but not if some have been removed. Alternatively we can delete the table's content and then just fill it up with the recently fetched list of bad hashes.

bad-bits/lib/badbits.js
NikolasHaimerl: Resolved in dec8653

bad-bits/bin/scheduler.js
juliangruber: That's not what I mean no. With the code being as it is, when we look at the cloudflare worker runs, we won't see a single failure. If we simply throw here, as suggested by me, we will see which runs failed. Logging helps in addition

NikolasHaimerl: I see, I don't see any issue with throwing the error here, so I'll go ahead and do that.

bad-bits/wrangler.toml
```diff
@@ -0,0 +1,30 @@
...
+ENVIRONMENT = "calibration"
+SECRET_HEADER_KEY = "Authorization"
+
+[[env.calibration.d1_databases]]
+binding = "DB"
+database_name = "filcdn-db"
+database_id = "8cc92155-16f6-426a-b782-2965e0daf103"
+
+[env.calibration.triggers]
+crons = ["*/5 * * * *"]  # Run every 5 minutes
```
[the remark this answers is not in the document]
juliangruber: @bajtos was there a particular reason to update this frequently?

bad-bits/bin/scheduler.js
juliangruber: It's not just about whether throwing an error creates an issue, it's also about whether not throwing it creates another one.

bad-bits/lib/store.js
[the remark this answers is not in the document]
juliangruber: > When it comes to comparing these hashes with the once from theindexer_roots table, I planned on checking them in batches, say 1000-10000 hashes at a time, similarly to how we check for [payload CIDs in spark deal observer](https://github.com/CheckerNetwork/spark-deal-observer/blob/main/backend/bin/deal-observer-backend.js#L132).

Can't we do that in one SQL query? Letting SQL handle the comparison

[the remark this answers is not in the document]
juliangruber: https://github.com/filcdn/worker/pull/145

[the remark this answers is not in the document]
juliangruber: > If we were to do all on one SQLite query it can potentially result in Connection timeouts, Memory overflows and things like SQL string too long due to the number of entries that are injected in the SQL query string being too big.

We are already inserting all hashes in one batch (eg when the DB is empty), I don't see how updating all hashes in one batch changes anything. And where are memory overflows possible?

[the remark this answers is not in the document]
juliangruber: Please describe where risk of memory overflows is introduced

juliangruber:
hmm there is another test error now

bad-bits/lib/store.js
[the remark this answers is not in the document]
NikolasHaimerl: - The worker only has 128MB in memory
- With a high number of hashes that need to be compared the parser and planner overhead of SQLite could lead to memory overflows. 
- SQLite also holds temporary values for the comparison query such as indices,  merge tables,... which can lead to memory overflows.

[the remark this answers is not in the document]
juliangruber: A memory overflow means there's a bug in the code leading to undefined behavior, see https://en.wikipedia.org/wiki/Buffer_overflow. Are you saying you're suspecting a memory overflow bug in SQLite? Or are you concerned about it simply using a lot of memory, and the worker running out of available memory?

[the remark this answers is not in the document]
NikolasHaimerl: I am concerned about it using more memory than the worker has available. 128MB is not that much and If we want to use this code in a mainnet version with possibly millions of hashes then we'll likely run into OOM issues if we do all in one SQLite query. This is why I proposed to do it in iterations so we can control how much is loaded in memory at a time.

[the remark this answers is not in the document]
NikolasHaimerl: Alright, I'll include an analysis in a follow up PR where the hashes are compared.

bad-bits/wrangler.toml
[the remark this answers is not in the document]
bajtos: Yes, it was suggested by Lidel. When new content is added to the badbits list, we want to start blocking retrievals of that file ASAP.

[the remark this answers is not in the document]
bajtos: I see how 5 minutes could be too often, how about refreshing the list every hour?

bad-bits/lib/store.js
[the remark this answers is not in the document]
bajtos: The helper `getAllBadBitHashes` was moved to test utils, I am closing this thread as resolved.

bad-bits/wrangler.toml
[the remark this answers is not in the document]
bajtos: I implemented etag-based caching. When the bad-bit list was not changed, our check is very quick - the server responds with 304 and an empty body. I think it's fine to run such check every 5 minutes.

bad-bits/bin/scheduler.js
```diff
@@ -0,0 +1,14 @@
+import { fetchAndStoreBadBits } from '../lib/bad-bits.js'
+
+export default {
+  async scheduled(_controller, env, _ctx) {
+    console.log('Running scheduled badbits update...')
```
juliangruber, line 5:
```suggestion
    console.log('Running scheduled bad bits update...')
```

bad-bits/bin/scheduler.js
```diff
@@ -0,0 +1,14 @@
...
+    try {
+      await fetchAndStoreBadBits(env)
+      console.log('Updated badbits denylist')
```
juliangruber, line 8:
```suggestion
      console.log('Updated bad bits denylist')
```

bad-bits/bin/scheduler.js
```diff
@@ -0,0 +1,14 @@
...
+    } catch (error) {
+      console.error('Failed to update badbits denylist:', error)
```
juliangruber, line 10:
```suggestion
      console.error('Failed to update bad bits denylist:', error)
```

bad-bits/lib/store.js
```diff
@@ -0,0 +1,53 @@
...
+      statements.push(
+        env.DB.prepare(
+          'INSERT INTO bad_bits_history (timestamp, etag) VALUES (?, ?)',
+        ).bind(now.toISOString(), etag),
+      )
+    }
+
+    await env.DB.batch(statements)
+  } catch (error) {
+    console.error('Error updating bad_bits:', error)
```
juliangruber, line 39:
```suggestion
    console.error('Error updating bad bits:', error)
```

PR #143, opened 2025-07-04
feat: store list of sanctioned addresses

NikolasHaimerl:
This PR adds sanctioned address checking to FilCDN. 

A new worker is added which checks all the addresses in the payment rail table (payer and payee). 
A new table is added for address statuses. These statuses can be (sanctioned, approved, pending). 
- Sanctioned: There exists a sanction for this address
- Approved: There does not exist a sanction for this address
- Pending: No check for this address has been made. 

Once a day, the addresses in the table containing the sanction status for each address (`address_sanction_check`) is compared with all payers and payee addresses. If the `address_sanction_check` is missing one of these addresses, they are copied over, and the status is set to `pending`. After that, all addresses in `address_sanction_check` with `pending` status are checked against the Chainalysis API. At maximum, per day, 1000 addresses are checked so that the Chainalysis API is not overloaded. 

Related to: 
https://github.com/filcdn/roadmap/issues/17

Considerations: 
At the current time, once an address has been marked as approved, they are no longer checked. This is done to avoid the overuse of the free Chainalysis API. Once we have a contract or agreement with Chainalysis in place for more requests, we can also check all approved addresses every time we run the worker.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -8,7 +8,6 @@ jobs:
     runs-on: ubuntu-latest
     env:
       CHAINALYSIS_API_KEY: ${{ secrets.CHAINALYSIS_API_KEY }}
-      GITHUB_ACTION: true
     steps:
       - uses: actions/checkout@v4
       - uses: actions/setup-node@v4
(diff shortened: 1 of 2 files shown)


pyropy:
I'm going to close this PR as it has been superseded by #189. However we can reuse portions of this code in the future to check if user has been sanctioned in the meantime.

PR #145, opened 2025-07-07
Refactor: compare in sql

juliangruber:
Move the badbits comparison to sql, speeding up and simplifying code. As a downside, the update stats have been removed. I doubt these are high value rn.


juliangruber:
@NikolasHaimerl please merge yourself if you agree with this approach, since this is targeting your PR

PR #155, opened 2025-07-09
feat: content-security-policy to prevent phishing

bajtos:
Enable Content Security Policy (CSP) in CDN responses to prevent malicious users from using our CDN for phishing. 

CSP is a security feature that helps prevent attacks like Cross-Site Scripting (XSS) by specifying which sources of content are allowed to be loaded by the browser.

The goal of this pull request is to configure the CDN response in such a way that browsers will not allow the page to interact with other services & websites.

In a nutshell:
- Allow self, inline, eval, blob, data and white-listed hosts as a source for JS, CSS, images, etc.
- Disable navigation or form submissions to other pages

List of allowed hosts - I expect us to evolve this list over time based on user feedback.
```js
const ALLOWED_HOSTS = [
  'https://*.filcdn.io',
  'https://*.calibration.filcdn.io',

  // Other services serving content-addressable or static assets
  'https://*.w3s.link',
  'https://*.dweb.link',
  'https://*.githubusercontent.com',
]
```

I decided to leave out reporting of CSP violations from this pull request and created a follow-up issue here:
- https://github.com/filcdn/worker/issues/156

Links:
- https://github.com/filcdn/roadmap/issues/31
- https://github.com/storacha/w3link/blob/d73e3783c4c520e85e96dba1a2eb507da0f3cbb3/packages/edge-gateway-link/src/gateway.js#L74-L98

### DIFF
--- a/retriever/bin/retriever.js
+++ b/retriever/bin/retriever.js
@@ -84,7 +84,7 @@ export default {
     const spURL =
       OWNER_TO_RETRIEVAL_URL_MAPPING[ownerAddress]?.url ||
       (await getProviderUrl(ownerAddress, env))
-    const { response, cacheMiss } = await retrieveFile(
+    const { response: originResponse, cacheMiss } = await retrieveFile(
       spURL,
       rootCid,
       env.CACHE_TTL,
@@ -96,7 +96,7 @@ export default {
       clientAddress: clientWalletAddress,
       cacheMiss,
       egressBytes: null, // Will be populated later
-      responseStatus: response.status,
+      responseStatus: originResponse.status,
       timestamp: requestTimestamp,
       performanceStats: {
         fetchTtfb: null, // Will be populated later
@@ -106,7 +106,7 @@ export default {
       requestCountryCode,
     }
 
-    if (!response.body) {
+    if (!originResponse.body) {
       // The upstream response does not have any readable body
       // There is no need to measure response body size, we can
       // return the original response object.
@@ -122,14 +122,14 @@ export default {
           },
         }),
       )
-      const clonedResponse = new Response(response.body, response)
-      setContentSecurityPolicy(clonedResponse)
-      return clonedResponse
+      const response = new Response(originResponse.body, originResponse)
+      setContentSecurityPolicy(response)
+      return response
     }
 
     // Stream and count bytes
     // We create two identical streams, one for the egress measurement and the other for returning the response as soon as possible
-    const [returnedStream, egressMeasurementStream] = response.body.tee()
+    const [returnedStream, egressMeasurementStream] = originResponse.body.tee()
     const reader = egressMeasurementStream.getReader()
     const firstByteAt = performance.now()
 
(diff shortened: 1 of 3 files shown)


retriever/lib/content-security-policy.js
```diff
@@ -0,0 +1,45 @@
+// List of allowed hosts in the CSP <host-source> format:
+// https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Content-Security-Policy#host-source
+const ALLOWED_HOSTS = [
+  'https://*.filcdn.io',
+  'https://*.calibration.filcdn.io',
```
juliangruber, line 5: I believe the above string covers that one as well
```suggestion
```

retriever/bin/retriever.js
```diff
@@ -121,7 +122,9 @@ export default {
           },
         }),
       )
-      return response
+      const clonedResponse = new Response(response.body, response)
```
juliangruber, line 125: can we rename to `writableResponse`, to clarify the intend?

retriever/lib/content-security-policy.js
```diff
@@ -0,0 +1,45 @@
...
+  // This functions sets the Content Security Policy (CSP) header for the response.
+  // CSP is a security feature that helps prevent attacks like Cross-Site Scripting (XSS) by specifying which sources of content are allowed to be loaded by the browser.
+  // Learn more:
+  //   https://developer.mozilla.org/en-US/docs/Web/HTTP/CSP
+  //   https://developer.mozilla.org/en-US/docs/Web/HTTP/Guides/CSP
+  //
+  // Our implementation is inspired by w3s.link:
+  // https://github.com/storacha/w3link/blob/d73e3783c4c520e85e96dba1a2eb507da0f3cbb3/packages/edge-gateway-link/src/gateway.js#L74-L98
+
+  const allowedHostsAsString = ALLOWED_HOSTS.join(' ')
```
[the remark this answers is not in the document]
juliangruber: I prefer the current way as the stringification is an implementation detail of this function, and we aren't yet micro optimizing

retriever/bin/retriever.js
juliangruber: sure, if we also update the comment in content-security-policy.js, which says it needs to be writable

bajtos: After trying different names, I decided to use a different approach:

- Rename to `fetch` response to `originResponse`
- Use just `response` for the Response objected returned by the worker

retriever/lib/content-security-policy.js
bajtos: I am not sure, TBH. Let's try your proposal; we can easily add back the other entry later if needed.

bajtos: Addressed in [be2b331](https://github.com/filcdn/worker/pull/155/commits/be2b331b52f12de560d4dd679f12e7fac7fd7b03)

retriever/bin/retriever.js
```diff
@@ -105,7 +106,7 @@ export default {
       requestCountryCode,
     }
 
-    if (!response.body) {
+    if (!originResponse.body) {
       // The upstream response does not have any readable body
```
bajtos, line 110: Instead of `originResponse`, we can also use `upstreamResponse`. WDYT?

juliangruber (approved):

PR #159, opened 2025-07-10
feat: send X-Origin-URL in response headers

bajtos:
Let the clients (like our bot) know what is the URL where the content can be retrieved from directly.

While working on https://github.com/filcdn/bot/pull/16, I realised it would be much easier for the bot if the CDN reported back the full piece retrieval URL it used (our would use) for cache-miss requests.

I asked GPT-4.1 about security concerns of exposing such URL. The only meaningful concern is that we are making it easier for attackers to retrieve from the SP directly, bypassing our CDN. I think that's not a problem, because the information for constructing the origin URL is public and can be easily found by making a few RPC API calls.

_This pull request includes changes from https://github.com/filcdn/worker/pull/158. I'll clean it up once the other PR is landed._


retriever/bin/retriever.js
```diff
@@ -124,6 +123,7 @@ export default {
       )
       const response = new Response(originResponse.body, originResponse)
       setContentSecurityPolicy(response)
+      setOriginUrl(response, originUrl)
```
bajtos, line 126: Hmm, maybe this name is a bit confusing. Would the following be easier to understand?


```js
setOriginUrlHeader(response, originUrl)
```

juliangruber: Or let's just inline this one-liner? That's simplest to me

juliangruber (commented):
Approach lgtm. I share your security intuition, especially if it's just for debugging. I was wondering, whether we should log this information in the CF retrieval worker instead?

bajtos:
Closing in favour of https://github.com/filcdn/worker/pull/163

Issue #160, opened 2025-07-10
Provider lookup must ignore providers not approved with Pandora


bajtos:
Consider the following retrieval request:

- ProofSet 442 
- Root 0 
- SP polynomial.computer 
- CDN request URL:
  https://0xce340d9a71b2aa8f7faa2f989158f14bede3e1b2.calibration.filcdn.io/baga6ea4seaqnbpfza3wl5fgu7gnfcyh5h6zufn4skwt6clylnzaw5k6maiwt6ay

This CommP is stored with two providers:

<img width="1086" height="146" alt="Image" src="https://github.com/user-attachments/assets/1c782014-931e-4c88-8d84-0b716df0fb25" />

The provider `0x12191de399B9B3FfEB562861f9eD62ea8da18AE5` was removed from the set of approved SPs because they are not fully operational.

Unfortunately, the logic in our retriever worker picks this SP as the origin for the cache-miss request.

We need to modify the logic to account for which SPs are approved.

Related:
- https://github.com/filcdn/roadmap/issues/13
- https://github.com/filcdn/worker/pull/107 removed `0x12191de399B9B3FfEB562861f9eD62ea8da18AE5` (`https://techx-pdp.filecoin.no`) from the list of supported SPs

bajtos:
Remaining tasks:

- Finish and deploy https://github.com/filcdn/worker/pull/163
- Finish https://github.com/filcdn/bot/pull/16
  - Address https://github.com/filcdn/bot/pull/16#discussion_r2197363586
  - Read X-ProofSet-ID response header and print it in the alert message

PR #163, opened 2025-07-11
feat: add X-Proof-Set-ID to response headers

bajtos:
- **test: fix BigInt format used by test data builders**

  Before this change, test data will store BigInt values formatted with one decimal place, e.g. `123.0`.

  This commit fixes the builders to correctly store BigInt values with no decimal places, e.g. `123`.

- **feat: add X-Proof-Set-ID to response headers**

  Let the clients (like our bot) know which ProofSet was chosen by the routing layer. This helps when troubleshooting failing retrievals in the case where the client stores the same content with multiple SPs.

Links:
- based on the discussion in https://github.com/filcdn/worker/pull/159
- https://github.com/filcdn/worker/issues/160


retriever/bin/retriever.js
```diff
@@ -67,11 +67,8 @@ export default {
     // Timestamp to measure file retrieval performance (from cache and from SP)
     const fetchStartedAt = performance.now()
 
-    const { ownerAddress, pieceRetrievalUrl } = await getOwnerAndValidateClient(
-      env,
-      clientWalletAddress,
-      rootCid,
-    )
+    const { ownerAddress, pieceRetrievalUrl, setId } =
```
juliangruber, line 70:
```suggestion
    const { ownerAddress, pieceRetrievalUrl, proofSetId } =
```

retriever/bin/retriever.js
```diff
@@ -119,6 +116,7 @@ export default {
       )
       const response = new Response(originResponse.body, originResponse)
       setContentSecurityPolicy(response)
+      response.headers.set('X-Dataset-ID', setId)
```
juliangruber, line 119: Has FilOz definitely decided on this terminology? I see a risk of making things more confusing if we integrate this new name too early (and nothing else is using it yet)

retriever/bin/retriever.js
bajtos: Considering the migration from "ProofSetId" to "DatasetId", I find the name "setId" better, because it works for both "ProofSetId" and "DatasetId".

juliangruber: it works ok for both but great of neither. `setId` can also refer to other things

bajtos: Fair enough, I'll rename to `proofSetId`.

retriever/bin/retriever.js
bajtos: See [d229586](https://github.com/filcdn/worker/pull/163/commits/d229586d5b8776352863905503b2230c2e00a4a4)


I decided to use `X-Proof-Set-ID` because "proofset" is not a word.


Screenshot 2025-07-16 at 11 41 53

retriever/bin/retriever.js
bajtos: Resolved in [d229586](https://github.com/filcdn/worker/pull/163/commits/d229586d5b8776352863905503b2230c2e00a4a4)

PR #164, opened 2025-07-11
fix: avoid double slash in piece retrieval URLs

bajtos:
Before this change, when the provider's URL ended with `/`, we constructed an URL with double slash in the path, e.g.

    https://polynomial.computer//piece/baga123

Curio does not support such paths and returns 404.

In this patch, I fix the code building the URL to handle this edge case and construct a correct URL with a single slash, e.g.

    https://polynomial.computer/piece/baga123

Links:
- https://github.com/filcdn/roadmap/issues/13

### DIFF
--- a/retriever/lib/retrieval.js
+++ b/retriever/lib/retrieval.js
@@ -21,7 +21,7 @@ export async function retrieveFile(
   cacheTtl = 86400,
   { signal } = {},
 ) {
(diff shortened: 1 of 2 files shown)


bajtos:
I am going to land this ASAP because our CDN is essentially broken now.

@juliangruber PTAL, I'll address review comments in a follow-up pull request.

juliangruber:
lgtm!

PR #168, opened 2025-07-11
Add naive HEAD request support

juliangruber:
Trying to use https://speedvitals.com/ttfb-test, I noticed that the CDN doesn't support HEAD requests yet.

To start off with a very simple and problem-free implementation (that even includes the reponse body, which is not against the spec), simply treat HEAD requests like GET requests.

I'm going to merge this before review, to unblock TTFB monitoring. Please hindsight review 🙏


bajtos:
LGTM 👍🏻

PR #171, opened 2025-07-15
feat: Check if payer address is sanctioned during indexing

pyropy:
This PR builds on the approach introduced in #143 by adding a check to determine if the payer address is sanctioned during the indexing of the proof-set rail creation.

While the checks from #143 run every 24 hours, they leave a window during which sanctioned addresses might perform retrievals before being flagged. This PR addresses that gap by introducing an check at indexing time, triggered only when the `with_cdn` flag is enabled for the proof-set rail.

Related to:
- https://github.com/filcdn/roadmap/issues/17
- #143

### DIFF
--- a/indexer/bin/indexer.js
+++ b/indexer/bin/indexer.js
@@ -164,10 +164,13 @@ export default {
       console.log(
         `New proof set rail (proof_set_id=${payload.proof_set_id}, rail_id=${payload.rail_id}, payer=${payload.payer}, payee=${payload.payee}, with_cdn=${payload.with_cdn})`,
       )
-      const isPayerSanctioned = await isAddressSanctioned(
-        CHAINALYSIS_API_KEY,
-        payload.payer,
-      )
+      let isPayerSanctioned
+      if (payload.with_cdn) {
+        isPayerSanctioned = await isAddressSanctioned(
+          CHAINALYSIS_API_KEY,
+          payload.payer,
+        )
+      }
       await env.DB.prepare(
         `
           INSERT INTO indexer_proof_set_rails (
@@ -188,7 +191,7 @@ export default {
           payload.payer,
           payload.payee,
           payload.with_cdn ?? null,
-          isPayerSanctioned,
+          isPayerSanctioned ?? null,
         )
         .run()
       return new Response('OK', { status: 200 })
--- a/indexer/bin/indexer.js
+++ b/indexer/bin/indexer.js
@@ -3,6 +3,7 @@ import {
   handleProviderRemoved,
 } from '../lib/provider-events-handler.js'
 import { createPdpVerifierClient as defaultCreatePdpVerifierClient } from '../lib/pdp-verifier.js'
+import { isAddressSanctioned as defaultIsAddressSanctioned } from '../lib/chainalysis.js'
 
 export default {
   /**
@@ -11,13 +12,17 @@ export default {
    * @param {ExecutionContext} ctx
    * @param {object} options
    * @param {typeof defaultCreatePdpVerifierClient} [options.createPdpVerifierClient]
+   * @param {typeof defaultIsAddressSanctioned} [options.isAddressSanctioned]
    * @returns {Promise<Response>}
    */
   async fetch(
     request,
     env,
     ctx,
-    { createPdpVerifierClient = defaultCreatePdpVerifierClient } = {},
+    {
+      createPdpVerifierClient = defaultCreatePdpVerifierClient,
+      isAddressSanctioned = defaultIsAddressSanctioned,
+    } = {},
   ) {
     // TypeScript setup is broken in our monorepo
     // There are multiple global Env interfaces defined (one per worker),
@@ -34,6 +39,8 @@ export default {
       SECRET_HEADER_KEY,
       // @ts-ignore
       SECRET_HEADER_VALUE,
+      // @ts-ignore
+      CHAINALYSIS_API_KEY,
     } = env
     if (request.headers.get(SECRET_HEADER_KEY) !== SECRET_HEADER_VALUE) {
       return new Response('Unauthorized', { status: 401 })
@@ -157,6 +164,10 @@ export default {
       console.log(
         `New proof set rail (proof_set_id=${payload.proof_set_id}, rail_id=${payload.rail_id}, payer=${payload.payer}, payee=${payload.payee}, with_cdn=${payload.with_cdn})`,
       )
+      const isPayerSanctioned = await isAddressSanctioned(
+        CHAINALYSIS_API_KEY,
+        payload.payer,
+      )
       await env.DB.prepare(
         `
           INSERT INTO indexer_proof_set_rails (
@@ -164,9 +175,10 @@ export default {
             rail_id,
             payer,
             payee,
-            with_cdn
+            with_cdn,
+            is_payer_sanctioned
           )
-          VALUES (?, ?, ?, ?, ?)
+          VALUES (?, ?, ?, ?, ?, ?)
           ON CONFLICT DO NOTHING
         `,
       )
@@ -176,6 +188,7 @@ export default {
           payload.payer,
           payload.payee,
           payload.with_cdn ?? null,
+          isPayerSanctioned,
         )
         .run()
       return new Response('OK', { status: 200 })
--- a/indexer/lib/chainalysis.js
+++ b/indexer/lib/chainalysis.js
@@ -1,0 +1,43 @@
+/**
+ * Check a single address with the Chainalysis API
+ *
+ * @param {string} apiKey - Chainalysis API key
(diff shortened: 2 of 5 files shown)


indexer/bin/indexer.js
```diff
@@ -157,16 +164,24 @@ export default {
       console.log(
         `New proof set rail (proof_set_id=${payload.proof_set_id}, rail_id=${payload.rail_id}, payer=${payload.payer}, payee=${payload.payee}, with_cdn=${payload.with_cdn})`,
       )
+      let isPayerSanctioned
+      if (payload.with_cdn) {
+        isPayerSanctioned = await isAddressSanctioned(
+          CHAINALYSIS_API_KEY,
+          payload.payer,
+        )
+      }
```
juliangruber, line 173: this can create slightly confusing data in the db, as we can store rows with `with_cdn=false` and `is_payer_sanctioned=false` even though the payer is sanctioned. For now, for correctness, can we always look up whether they are sanctioned?

indexer/lib/chainalysis.js
```diff
@@ -0,0 +1,43 @@
...
+export async function isAddressSanctioned(
+  apiKey,
+  address,
+  { fetch = globalThis.fetch } = {},
+) {
+  try {
+    const url = `https://public.chainalysis.com/api/v1/address/${address}`
+
+    const response = await fetch(url, {
+      method: 'GET',
```
juliangruber, line 19:
```suggestion
```

this is the default

indexer/lib/chainalysis.js
```diff
@@ -0,0 +1,43 @@
...
+        Accept: 'application/json',
+      },
+    })
+
+    if (!response.ok) {
+      console.error(
+        `Chainalysis API error: ${response.status} ${response.statusText}`,
+      )
+
+      return null
```
juliangruber, line 31: In this case, we can store a sanctioned payer as not sanctioned, which is risky. I think we have to retry, until the chainalysis call worked. Ideally, we create a queue for this, so retries are reliable. I'm not sure what the best approach for M2 is though 🤔 We also don't want to not index proof sets, when chainalysis is unavailable, and the payer is not sanctioned.

@bajtos wdyt?

juliangruber (changes requested):
great start!

indexer/bin/indexer.js
pyropy: In case `with_cdn = false` we won't check if address is sanctioned and it should default to `null`. I can see how the data can be confusing in that case but I've done this only to save some API calls towards Chainalysis.

indexer/lib/chainalysis.js
pyropy: There seems to be a [Cloudflare Queue](https://developers.cloudflare.com/queues/configuration/batching-retries/) product that could be integrated together with the Cloudflare Worker to handle retries.

indexer/bin/indexer.js
juliangruber: good point, I didn't see that it would be `null` and not `false`. Wfm!

indexer/lib/chainalysis.js
pyropy: I have added the retry process in this commit https://github.com/filcdn/worker/pull/171/commits/13e6e870e23337a9d54601607d60de9d76bd28f3. Maybe it would be a good idea to add some limits on how many times the message should be retried before we drop it or send it to some other queue?

bajtos: +1 to add a retry queue 👍🏻

bajtos (dismissed):
It's a great idea to check the payer address during indexing.

I feel it's important to discuss several higher-level aspects before I look into implementation details.

1. A single client (payer) can have more than one proofset (more than one FS deal). In the current proposal, when the client creates a new proofset, we don't take into account their existing proofsets and do a new chainanalysis lookup. I guess that's fine, as it makes it more likely that we detect faster when the address became flagged.

2. However, if an address was flagged by Chainalysis, we update only the latest proofset. This creates inconsistent state, where some proofset of the same address as flagged as sanctioned but others are not.

   Suggestion: create a new table `(wallet_address, is_sanctioned)`.

3. An edge case to consider: there are two proofsets from the same payer created in the same Filecoin epoch (block).

4. What is the UX when the Chainanalysis check is pending? The user gets a retrieval URL from the Synapse SDK and sends a retrieval request - what message will they receive back? Does the message explain what's going on and what should the user do?   In an ideal world, I would like the message to tell the user that we are checking their address and they should retry the retrieval in a minute. A more generic message can be sufficient for M2, if it's easier to implement.

5. I think we should also check whether the SP (`payee`) is sanctioned. While this is a post-M2 feature, we may want to design the database schema to make it easier to implement payee checks later.

db/migrations/0009_add_wallet_details_table.sql
```diff
@@ -0,0 +1,13 @@
+CREATE TABLE IF NOT EXISTS wallet_details (
+  wallet_address TEXT PRIMARY KEY,
+  is_sanctioned BOOLEAN NOT NULL DEFAULT FALSE
+);
```
bajtos, line 4: I am wondering if we should use the same data schema as was proposed in https://github.com/filcdn/worker/pull/143/files#diff-9ae133790f3d9c9aabc53407feba0ca412ba43527ceb8cf3cb5a1896ddcc73dc?


```sql
CREATE TABLE IF NOT EXISTS wallet_details (
  wallet_address TEXT PRIMARY KEY,
  sanction_check_status TEXT NOT NULL DEFAULT 'pending'
    CHECK (sanction_check_status IN ('sanctioned', 'approved', 'pending'))
);
```

indexer/lib/proof-set-handler.js
```diff
@@ -20,16 +20,25 @@ export async function handleProofSetRailCreated(
...
+        INSERT INTO wallet_details (wallet_address, is_sanctioned)
+        VALUES (?, ?)
+        ON CONFLICT (wallet_address) DO UPDATE SET is_sanctioned = excluded.is_sanctioned
+      `,
+      )
+        .bind(payload.payer, isPayerSanctioned)
+        .run()
+    } catch (err) {
+      console.error(`Error checking if payer is sanctioned: ${err}`)
+      throw err // Let caller handle the error
```
bajtos, line 40: If we are just re-throwing the error to let the caller handle it, do we need the try/catch block at all?

bajtos (commented):
Great progress!

db/migrations/0009_add_wallet_details_table.sql
bajtos: Maybe we can set `is_sanctioned = NULL` to indicate that we haven't checked the sanction status yet?

I think we can also drop the `wallet_` prefix from the column name, since the table name `wallet_details` already indicates that the address relates to wallets.

Just ideas to consider, please decide for yourself which of these suggestions make sense and which to ignore (or defer for later).

indexer/bin/indexer.js
```diff
@@ -189,4 +183,34 @@ export default {
...
+   *
+   * @param {MessageBatch<{ type: string; payload: any }>} batch
+   * @param {Env} env
+   * @param {object} options
+   * @param {typeof defaultIsAddressSanctioned} [options.isAddressSanctioned]
+   */
+  async queue(
+    batch,
+    env,
+    { isAddressSanctioned = defaultIsAddressSanctioned } = {},
```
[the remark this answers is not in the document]
pyropy: I've changed it to `checkIfAddressIsSanctioned` in https://github.com/filcdn/worker/pull/171/commits/4ec2fe56c413aea782b57e0be3407ec55bdc6eb3

db/migrations/0009_add_wallet_details_table.sql
pyropy: Changed column name to `address` in https://github.com/filcdn/worker/pull/171/commits/4ec2fe56c413aea782b57e0be3407ec55bdc6eb3

db/migrations/0009_add_wallet_details_table.sql
```diff
@@ -0,0 +1,19 @@
+CREATE TABLE IF NOT EXISTS wallet_details (
+  address TEXT PRIMARY KEY,
+  is_sanctioned BOOLEAN
+);
+
+-- Backfill existing wallet addresses
```
juliangruber, line 6: This shouldn't be necessary, as also need to insert rows when we first check their sanctions status. But, it also doesn't hurt

juliangruber (approved):
Lgtm! Shall we include the matching retriever logic in this PR too? Or is that better kept in a separate one?

db/migrations/0009_add_wallet_details_table.sql
pyropy: I should include that into a separate PR.

PR #176, opened 2025-07-16
feat: batch bad-bits SQL updates by 1000 items

bajtos:
The bad-bits updater worker fails with Exceeded Resources.

In this pull request, I am reworking the code executing SQL update queries to execute updates in batches of 1000 items.

### DIFF
--- a/bad-bits/lib/store.js
+++ b/bad-bits/lib/store.js
@@ -7,38 +7,38 @@
  * @param {string} etag - ETag for the current denylist
  */
 export async function updateBadBitsDatabase(env, currentHashes, etag) {
-  try {
-    const now = new Date()
-    const insertBadBitStmt = env.DB.prepare(
-      `
+  const now = new Date().toISOString()
+  const insertBadBitStmt = env.DB.prepare(
+    `
       INSERT INTO bad_bits (hash, last_modified_at) VALUES (?, ?)
       ON CONFLICT(hash) DO UPDATE SET last_modified_at = excluded.last_modified_at
       `,
+  )
+
+  const remainingHashes = Array.from(currentHashes)
+  while (remainingHashes.length > 0) {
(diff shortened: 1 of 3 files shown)


bajtos:
With these changes deployed, we are hitting the 15-minute execution limit now.

Screenshot 2025-07-16 at 15 55 42

bajtos:
Batches of 10k items seem to hit the sweet spot - we update ~480k records in ~5 minutes.

bad-bits/lib/store.js
```diff
@@ -7,38 +7,42 @@
...
+  while (remainingHashes.length > 0) {
+    // pop first N hashes from remainingHashes
+    const batchHashes = remainingHashes.splice(0, 10_000)
+    updated += batchHashes.length
+    await env.DB.batch(
+      batchHashes.map((hash) => insertBadBitStmt.bind(hash, now)),
+    )
+    console.log(
+      `Inserted/updated ${updated} bad bits in ${Date.now() - startedAt}ms`,
+    )
```
bajtos, line 30: Should we keep this log for diagnostics, or is it just noise that's better to remove?

bad-bits/lib/store.js
```diff
@@ -7,38 +7,42 @@
  * @param {string} etag - ETag for the current denylist
  */
 export async function updateBadBitsDatabase(env, currentHashes, etag) {
-  try {
```
bajtos, line 10: I moved the try/catch block to `fetchAndStoreBadBits` as it makes the code in `updateBadBitsDatabase` easier to read, IMO.

bad-bits/lib/store.js
juliangruber: to me fine either way, happy to follow up if I later think it's noisy

juliangruber (approved):

PR #198, opened 2025-07-30
indexer: add goldsky monitoring

juliangruber:
This will alert us about Subgraph errors or indexing issues.

### DIFF
--- a/indexer/bin/indexer.js
+++ b/indexer/bin/indexer.js
@@ -221,4 +221,65 @@ export default {
       }
     }
   },
+
+  /**
+   * @param {any} _controller
+   * @param {Env} _env
+   * @param {ExecutionContext} _ctx
+   * @param {object} [options]
+   * @param {typeof globalThis.fetch} [options.fetch]
+   */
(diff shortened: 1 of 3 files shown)


pyropy (changes requested):
In the example found in [cloudflare docs](https://developers.cloudflare.com/workers/runtime-apis/handlers/scheduled/) they are using `ctx.waitUntil(promisePromise)` to run the function inside the `scheduled` handler. 

This is a quote from their docs

> Use this method to notify the runtime to wait for asynchronous tasks (for example, logging, analytics to third-party services, streaming and caching)...

I suppose we should do the same and wait for the promises to resolve.

juliangruber:
Interesting! We aren't using that for bad bits yet (https://github.com/filcdn/worker/blob/9534117bc139314cff0ada8ee2ce622b57cb1690/bad-bits/bin/scheduler.js) but the docs indicate we should. I don't understand why we should :D But I will go with the recommendation. Great find! Will also create a bad bits PR

PR #199, opened 2025-07-31
retriever: fix await execution context in tests

juliangruber:
There is a clouflare:test native way to do this, no need to write a custom worker wrapper


pyropy (approved):
Great find & great work! 🚀

Issue #225, opened 2025-08-21
Ignore retrieval requests for terminated rails


bajtos:
- Index rail termination events
- Retriever should ignore data sets with terminated rails

Ideally, the smart contract should clear the withCDN flag when the CDN payment rail is terminated and emit an event that we can listen for.

See https://docs.google.com/document/d/1_X9_cQ6kA99o_hdm0fn34atKPBvbobOAzWU-f-aoh0g/edit?tab=t.0 and https://www.notion.so/spacemeridian/Sanction-screening-of-FilCDN-clients-256cdd5cccdb80c9be5efa3a42d6174f?source=copy_link for additional context.

Sub-tasks:
- [x] Setup Goldsky pipeline to call our webhook for relevant FWSS events
- [x] Implement webhook handlers for those events
- [x] Modify the retriever worker to return an error when the deal was flagged as sanctioned. 
   Note: We may be able to leverage the already existing sanction checks implemented by the retriever worker.

Note that we don't need to change filcdn-bot, because when the CDN service is terminated, the withCDN flag is cleared, and the bot is already skipping such deals.

PR #229, opened 2025-08-22
Handle storage and CDN service termination

pyropy:
Add handlers for storage and cdn service termination events to indexer. Once the service is terminated data set `with_cdn` property is set to `false` rather then deleting the data set.

Closes https://github.com/filcdn/worker/issues/225


indexer/lib/filecoin-warm-storage-service-handlers.js
```diff
@@ -59,3 +59,42 @@ export async function handleFilecoinWarmStorageServiceDataSetCreated(
     )
     .run()
 }
+
+/**
+ * @param {Env} env
+ * @param {any} payload
+ * @throws {Error}
+ */
+export async function handleFilecoinWarmStorageServiceStorageServiceTerminated(
```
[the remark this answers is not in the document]
juliangruber: +1 to the suggested name.

In some cases I'm mentioning the contract name in other work, because for example both FilecoinWarmStorageService and PDPVerifier will emit a `DataSetCreated` event.

I have not been using the abbreviation so far, I don't find it more readable.

juliangruber (commented):
@pyropy could you please request another review once we landed the PR blocking this one?

indexer/test/indexer.test.js
```diff
@@ -905,6 +954,28 @@ describe('POST /service-provider-registry/provider-removed', () => {
...
+  await env.DB.prepare(
+    `
+    INSERT INTO data_sets (
+      id,
+      with_cdn,
+      payer,
+      payee
+    )
+    VALUES (?, ?, ?, ?)
+    ON CONFLICT DO NOTHING
```
bajtos, line 970: This is risky; it can hide test failure causes, making it difficult to troubleshoot. It can also make a test to always pass because the test data created by the test was discarded, and slightly different data from previous tests was used instead.

I propose to:
- Remove this `ON CONFLICT` line, let the function throw an error on conflicts
- Add `beforeEach` hooks to reset the database before each test to avoid conflicts.

indexer/test/indexer.test.js
```diff
@@ -905,6 +954,28 @@ describe('POST /service-provider-registry/provider-removed', () => {
...
+  `,
+  )
+    .bind(String(dataSetId), withCDN ? 1 : 0, payer, payee)
```
bajtos, line 973: I think this can be simplified as follows:

```suggestion
    .bind(String(dataSetId), withCDN, payer, payee)
```

bajtos (changes requested):

bajtos (approved):

pyropy:
> @pyropy can you please pull in the latest `update/m2.5` version?

Done!

PR #243, opened 2025-09-03
feat: Terminate CDN service for sanctioned client

pyropy:
This pull request introduces the new `terminator` service to the monorepo, which automates the process of terminating CDN services for sanctioned wallets. It includes the core worker logic, queue and workflow handlers, contract interaction utilities, and corresponding tests. Additionally, the CI/CD pipeline and configuration files have been updated to support deployment and operation of the new service.

The new `terminator` service will periodically scan for the sanctioned wallets and send requests for (CDN) service termination to the queue. Once when the request is read from the queue and transaction is sent worker will spawn a `TransactionMonitorWorkflow` that will wait for transaction execution.  

In case the transaction execution takes too long transaction `TransactionMonitorWorkflow` will send a transaction cancelation request to the queue. Once the request is read from the queue a new cancel transaction is sent by the worker and a new `TransactionMonitorWorkflow` is spawned in order to monitor the new cancel transaction.

Closes https://github.com/filcdn/roadmap/issues/71


pyropy:
>  What happens after we cancel all outstanding services for a sanctioned address?

Once the service is cancelled termination events should be emitted and handled by handlers introduced in https://github.com/filcdn/worker/pull/229. 

> * The next scheduled run (periodic scan) should not call the termination smart contract method again, because all services were already terminated.

That's true. Currently we're not handling the case where termination has finished but it hasn't been indexed yet. In that case another call to the contract with be issued but smart contract will abort the call. 

> * When the sanctioned client creates a new service (a new dataset?), then we do need to call the termination method.

That's also true. Current logic does flag address as sanctioned during the data set creation but does not invoke termination call.

service-monitor/bin/service-monitor.js
```diff
@@ -0,0 +1,112 @@
...
+ *     | 'https://api.node.glif.io/'
+ *   FILECOIN_WARM_STORAGE_SERVICE_ADDRESS: string
+ *   FILCDN_CONTROLLER_ADDRESS_PRIVATE_KEY: string
+ *   DB: D1Database
+ *   TRANSACTION_QUEUE: import('cloudflare:workers').Queue<
+ *     TerminateServiceMessage | TransactionCancelMessage
+ *   >
+ *   TRANSACTION_MONITOR_WORKFLOW: import('cloudflare:workers').WorkflowEntrypoint
+ * }} ServiceMonitorEnv
+ *   }} ServiceMonitorEnv
```
bajtos, line 41: Typo?


```suggestion
 * }} ServiceMonitorEnv
```

service-monitor/lib/terminate-cdn-service.js
```diff
@@ -0,0 +1,33 @@
...
+ */
+export async function terminateCDNServiceForSanctionedWallets(env) {
+  const { results: dataSets } = await env.DB.prepare(
+    `
+      SELECT DISTINCT ds.id
+      FROM data_sets ds
+        LEFT JOIN wallet_details wd ON ds.payer_address = wd.address
+      WHERE ds.with_cdn = 1 AND wd.is_sanctioned = 1;
+  `,
+  ).run()
```
bajtos, line 17: It will take some time until the TXs terminating the datasets are accepted by the network, Goldsky picks up the termination events and invokes our webhook.

During that time, our scheduled worker will be called, possibly more than once. Those calls will pick up the same set of datasets to terminate, and create more transactions with duplicate "terminate service" calls for the same data set ID.

IMO, we need to keep track of which datasets have a pending termination TX, and ignore them in this query.

service-monitor/lib/terminate-cdn-service.js
```diff
@@ -0,0 +1,35 @@
...
+
+  console.log(`Found ${dataSets.length} sanctioned data sets to terminate`)
+
+  // Send messages to queue for processing
+  const messages = dataSets.map(({ id: dataSetId }) => ({
+    body: {
+      dataSetId,
+      type: 'terminate-cdn-service',
+    },
+  }))
```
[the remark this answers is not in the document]
pyropy: Yes that'd be nice to have but it should also work without it. I propose that we add that in a follow-up PR.

service-monitor/lib/terminate-cdn-service.js
pyropy: Yes that's true. While the CF Workflows can solve such problems using their unique IDs it's not possible to solve this problem using queues without persisting state in either durable object or database. 

Another (temporary) solution would be to have this worker triggered at lower frequency (i.e. every 10 minutes).

pyropy: > IMO, we need to keep track of which datasets have a pending termination TX, and ignore them in this query.

Added `terminate_service_tx_hash ` column to data sets in https://github.com/filcdn/worker/pull/243/commits/2b4c23c635a61bc9b8294ec4fa079f4895129834 and https://github.com/filcdn/worker/pull/243/commits/7b4681e98cbf37723bf2f21f98a5ff9a4f283582

PR #248, opened 2025-09-04
Update to service provider registry and beneficiary separation

juliangruber:
This should bring us up to date with the deployed state of FWSS contracts

### DIFF
--- a/retriever/test/retriever.test.js
+++ b/retriever/test/retriever.test.js
@@ -1,7 +1,7 @@
 import { describe, it, expect, vi, beforeAll } from 'vitest'
 import worker from '../bin/retriever.js'
-// import { createHash } from 'node:crypto'
-// import { retrieveFile } from '../lib/retrieval.js'
+import { createHash } from 'node:crypto'
+import { retrieveFile } from '../lib/retrieval.js'
 import {
   env,
   createExecutionContext,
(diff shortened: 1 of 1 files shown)


retriever/test/retriever.test.js
```diff
@@ -1,7 +1,7 @@
 import { describe, it, expect, vi, beforeAll } from 'vitest'
 import worker from '../bin/retriever.js'
-import { createHash } from 'node:crypto'
-import { retrieveFile } from '../lib/retrieval.js'
+// import { createHash } from 'node:crypto'
+// import { retrieveFile } from '../lib/retrieval.js'
```
bajtos, line 4: Commented-out code, can we remove it?

retriever/test/retriever.test.js
```diff
@@ -367,113 +363,111 @@ describe('retriever.fetch', () => {
...
-          (o) => o.storageProviderAddress,
-        )
-        throw new Error(
-          `❌ All storage providers failed to fetch. Storage providers checked: ${storageProvidersChecked.join(', ')}`,
-        )
-      }
-    },
-  )
+  // TODO: Enable test, one contest on calibration exists
+  // it(
```
bajtos, line 381: ditto, can we use `it.skip`?

bajtos (approved):
Awesome work, @juliangruber!

I skimmed through the changes, and they look reasonable.

My only concern is about removing the SP & client addresses from the retrieval logs table. I have a gut feeling that we will have to bring them back in the future, maybe as soon as we start updating the dashboard. However, I don't have any solid arguments for keeping them, and we can fill back old retrieva log rows from the data set table if needed, so I see this change as relatively easy to reverse.

retriever/test/retriever.test.js
juliangruber: 0f0c6b9

retriever/test/retriever.test.js
juliangruber: 0f0c6b9

db/migrations/0014_m2_5_upgrades.sql
```diff
@@ -5,20 +5,17 @@ DROP TABLE indexer_roots;
...
   service_url TEXT,
   PRIMARY KEY (id)
 );
-CREATE INDEX providers_beneficiary_address ON providers(beneficiary_address);
 
 CREATE TABLE data_sets (
   id TEXT NOT NULL,
-  storage_provider_address TEXT,
-  payer_address TEXT,
-  payee_address TEXT,
```
pyropy, line 20: Without `payee` and `storage_provider_address` we're unable to scan them again the sanctions API.

PR #261, opened 2025-09-08
Feat/terminate cdn service for sanctioned client

pyropy:
Reopens #243 that after accidental merge. 

This pull request introduces the new terminator service to the monorepo, which automates the process of terminating CDN services for sanctioned wallets. It includes the core worker logic, queue and workflow handlers, contract interaction utilities, and corresponding tests. Additionally, the CI/CD pipeline and configuration files have been updated to support deployment and operation of the new service.

The new terminator service will periodically scan for the sanctioned wallets and send requests for (CDN) service termination to the queue. Once when the request is read from the queue and transaction is sent worker will spawn a TransactionMonitorWorkflow that will wait for transaction execution.

In case the transaction execution takes too long transaction TransactionMonitorWorkflow will send a transaction cancelation request to the queue. Once the request is read from the queue a new cancel transaction is sent by the worker and a new TransactionMonitorWorkflow is spawned in order to monitor the new cancel transaction.

Closes https://github.com/filcdn/roadmap/issues/71

### DIFF
--- a/indexer/wrangler.toml
+++ b/indexer/wrangler.toml
@@ -70,22 +70,3 @@ binding = "RETRY_QUEUE"
 queue = "indexer-retry-queue-mainnet"
 max_batch_size = 100
 max_batch_timeout = 30
-
-[[env.calibration.queues.producers]]
-queue = "retry-queue-calibration"
-binding = "RETRY_QUEUE"
-
-[[env.calibration.queues.consumers]]
-queue = "retry-queue-calibration"
-max_batch_size = 100
-max_batch_timeout = 30
-
-[[env.mainnet.queues.producers]]
-queue = "retry-queue-mainnet"
-binding = "RETRY_QUEUE"
-
-[[env.mainnet.queues.consumers]]
-queue = "retry-queue-mainnet"
-max_batch_size = 100
-max_batch_timeout = 30
-


terminator/lib/terminate-cdn-service.js
```diff
@@ -0,0 +1,33 @@
...
+  // Send messages to queue for processing
+  const messages = dataSets.map(({ id: dataSetId }) => ({
+    body: {
+      dataSetId,
+      type: 'terminate-cdn-service',
+    },
+  }))
+
+  if (messages.length > 0) {
+    await env.TRANSACTION_QUEUE.sendBatch(messages)
```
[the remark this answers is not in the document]
juliangruber: otherwise, using the queue is smart and a stable solution 👏

terminator/lib/queue-handlers.js
```diff
@@ -0,0 +1,196 @@
...
+ *
+ * @param {TerminateServiceMessage} message
+ * @param {Env} env
+ */
+export async function handleTerminateServiceQueueMessage(
+  message,
+  env,
+  { getChainClient = defaultGetChainClient } = {},
+) {
+  const { dataSetId } = message
```
juliangruber, line 44:
```suggestion
  const { dataSetId } = message
  assert(dataSetId)
```

safety first

terminator/lib/transaction-monitor-workflow.js
```diff
@@ -0,0 +1,62 @@
...
+          })
+        },
+      )
+
+      await step.do(
+        'clean terminate_service_tx_hash when transaction is confirmed',
+        {},
+        async () => {
+          this.env.DB.prepare(
+            'UPDATE data_sets SET terminate_service_tx_hash = NULL WHERE terminate_service_tx_hash = ?',
```
juliangruber, line 42:
```suggestion
            'UPDATE data_sets SET terminate_service_tx_hash = NULL, with_cdn = 0 WHERE terminate_service_tx_hash = ?',
```

should we do this here as well to guarantee consistency?

terminator/lib/transaction-monitor-workflow.js
```diff
@@ -0,0 +1,62 @@
...
+        async () => {
+          const { publicClient } = getChainClient(this.env)
+          return await publicClient.waitForTransactionReceipt({
+            hash: transactionHash,
+          })
+        },
+      )
+
+      await step.do(
+        'clean terminate_service_tx_hash when transaction is confirmed',
```
juliangruber, line 38: To me it was surprising that we clear the hash after confirmation. Could we either keep it (for our record keeping) or rename the column to `pending_terminate_service_tx_hash`?

juliangruber: My preference would be to keep it.

terminator/lib/transaction-monitor-workflow.js
```diff
@@ -0,0 +1,62 @@
+/** @import {WorkflowEvent, WorkflowStep} from 'cloudflare:workers' */
+import { WorkflowEntrypoint } from 'cloudflare:workers'
+import { getChainClient as defaultGetChainClient } from './chain.js'
+
+/**
+ * Workflow that monitors a transaction and starts cancel workflow if it takes
+ * too long
+ */
+export class TransactionMonitorWorkflow extends WorkflowEntrypoint {
```
juliangruber, line 9: Awesome to use a monitor for this!

terminator/lib/queue-handlers.js
```diff
@@ -0,0 +1,196 @@
...
+ *
+ * @param {TransactionCancelMessage} message
+ * @param {Env} env
+ */
+export async function handleTransactionCancelQueueMessage(
+  message,
+  env,
+  { getChainClient = defaultGetChainClient } = {},
+) {
+  const { transactionHash } = message
```
juliangruber, line 110:
```suggestion
  const { transactionHash } = message
  assert(transactionHash)
```

terminator/lib/queue-handlers.js
```diff
@@ -0,0 +1,196 @@
...
+
+  console.log(
+    `Processing transaction cancellation for hash: ${transactionHash}`,
+  )
+
+  try {
+    const { publicClient, walletClient } = getChainClient(env)
+
+    // First check if the original transaction is still pending
```
juliangruber, line 119: 👏

terminator/lib/queue-handlers.js
```diff
@@ -0,0 +1,196 @@
...
+    // Get the original transaction to determine gas price
+    const originalTx = await publicClient.getTransaction({
+      hash: transactionHash,
+    })
+
+    console.log(
+      `Retrieved original transaction ${transactionHash} for cancellation`,
+    )
+
+    // Calculate higher gas price (150% of original)
```
juliangruber, line 147: I assume this logic was taken from the cancel stuck tx lib?

terminator/lib/queue-handlers.js
```diff
@@ -0,0 +1,196 @@
...
+
+    console.log(
+      `Sent cancellation transaction ${cancelHash} for original transaction ${transactionHash}`,
+    )
+
+    await env.DB.prepare(
+      `UPDATE data_sets SET terminate_service_tx_hash = ? WHERE terminate_service_tx_hash = ?`,
+    )
+      .bind(cancelHash, transactionHash)
+      .run()
```
juliangruber, line 176:
```suggestion
    await env.DB.prepare(
      `UPDATE data_sets SET terminate_service_tx_hash = ? WHERE terminate_service_tx_hash = ?`,
    )
      .bind(cancelHash, transactionHash)
      .run()
```

This is confusing, as now we're storing the hash of a cancel transaction - but nothing tells me as a viewer that this is a cancel transaction and not a terminate transaction

terminator/bin/terminator.js
```diff
@@ -0,0 +1,111 @@
...
+    } = {},
+  ) {
+    for (const message of batch.messages) {
+      console.log(
+        `Processing transaction queue message of type: ${message.type}`,
+      )
+      try {
+        switch (message.body.type) {
+          case 'terminate-cdn-service':
+            await handleTerminateServiceQueueMessage(message.body, env)
```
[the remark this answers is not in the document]
pyropy: Handled in https://github.com/filcdn/worker/pull/261/commits/feb391dd09acd93167ea23e49bc164c18cff8ba1

terminator/lib/terminate-cdn-service.js
```diff
@@ -0,0 +1,33 @@
...
+ *   }>
+ * }} env
+ */
+export async function terminateCDNServiceForSanctionedWallets(env) {
+  const { results: dataSets } = await env.DB.prepare(
+    `
+      SELECT DISTINCT ds.id
+      FROM data_sets ds
+        LEFT JOIN wallet_details wd ON ds.payer_address = wd.address
+      WHERE ds.with_cdn = 1 AND wd.is_sanctioned = 1 AND ds.terminate_service_tx_hash IS NULL;
```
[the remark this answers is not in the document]
bajtos: I also find it easier to read the query with full names (`wallet_details` instead of `wd`), but I don't feel too strongly about this.

terminator/lib/fwss.js
```diff
@@ -0,0 +1,15 @@
+export const abi = [
+  {
+    type: 'function',
+    name: 'terminateCDNService',
```
bajtos, line 4: In the past, we were using Solidity ABI syntax:
- https://github.com/filcdn/bot/blob/71414382d34437644a24b34f48135aebe3b176b3/index.js#L14-L39
- https://github.com/filcdn/worker/blob/1b8aed7cd7a5bae13b4234557c21c492faf172d7/indexer/lib/pdp-verifier.js#L5-L8 (we removed this as part of M2.5 upgrade)

I don't mind switching to JSON ABI as you do here, but I'd like to avoid inconsistency. I'd like us to either use JSON ABI everywhere, or use Solidity ABI everywhere, but not both.

bajtos (changes requested):
Let's discuss the handling of stuck transactions. I think the current proposal may not work well.

terminator/lib/fwss.js
juliangruber: Since we're using viem here, maybe it's different?

bajtos: Fair enough :+1:

If we are using Viem in our workers, should be eventually migrate the bot from Ethers to Viem too? (We can create a tech-debt issue in the bot repo, and ask PLDG to contribute.)

terminator/lib/queue-handlers.js
```diff
@@ -0,0 +1,198 @@
...
+    )
+
+    // Calculate higher gas price (150% of original)
+    const newGasPrice = originalTx.gasPrice
+      ? (originalTx.gasPrice * 3n) / 2n
+      : undefined
+    const newMaxFeePerGas = originalTx.maxFeePerGas
+      ? (originalTx.maxFeePerGas * 3n) / 2n
+      : undefined
+    const newMaxPriorityFeePerGas = originalTx.maxPriorityFeePerGas
```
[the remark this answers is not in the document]
pyropy: I just went with something really simple but I think that your approach is much better. I've updated the cancel transaction handler to try and retry the transaction rather then canceling it. It's now using formulas from the `cancel-struck-transaction` library you've mentioned.

terminator/lib/queue-handlers.js
pyropy: > Maybe we shouldn't cancel TXs, but resend the same TX with higher gas allowance.

Yes, I think that's actually a better approach. I've updated the handlers to retry the transaction rather then canceling it.

PR #262, opened 2025-09-08
feat: add subgraph for PDPVerifier Piece events

bajtos:
Introduce our own subgraph to transform the wire format of event properties
sent by Goldsky to our webhooks. This subgraph is indexing only PDPVerifier
events `PiecesAdded` and `PiecesRemoved`. We can add indexing of more events
and other smart contracts later.

A testing version of this subgraph is deployed at Goldsky here:
https://api.goldsky.com/api/public/project_cmb91qc80slyu01wca6e2eupl/subgraphs/mb-pdp-verifier-alpha-calibnet/0.1.0/gn

Links:
- https://github.com/filcdn/roadmap/issues/59

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,6 +1,9 @@
 name: CI
 on:
   push:
+    branches:
+      - main
+      - 'update/m2.5'
   pull_request:
 
 jobs:


bajtos:
Discussion points:
- Should this PR target the m2.5 branch to avoid merge conflicts in the package-lock file?
- I added the subgraph to the worker monorepo so that we can change the indexer worker and the subgraph defining webhook payloads together. Do you agree it's a good approach?
- There is no automated CD pipeline to push subgraph versions to Goldsky.

bajtos:
More discussion: as I discovered when I tried to run tests, `graph` tooling is not compatible with monorepos. We have to keep the subgraph project as a standalone npm project, decoupled from our monorepo :( We will need to configure Dependabot to look at `subgraph` in addition to the root.

bajtos:
>That's also fine with me, as long as it's documented. Back to my original question, can we revert any changes to the root package-lock.json then?

Yes, I hope so. Let me try.

bajtos:
The CI failures are unrelated to my changes.

Screenshot 2025-09-08 at 17 12 43

PR #264, opened 2025-09-09
fix: db deployment scripts

bajtos:
- fix the calibration DB name
- refactor the code - extract config into variables to avoid repetition

Output from the fixed script:

```
❯ ./db/deploy-calibration.sh

 ⛅️ wrangler 4.33.1 (update available 4.34.0)
─────────────────────────────────────────────
Migrations to be applied:
┌────────────────────────┐
│ name                   │
├────────────────────────┤
│ 0014_m2_5_upgrades.sql │
└────────────────────────┘
? About to apply 1 migration(s)
Your database may not be available to serve requests during the migration, continue? › (Y/n)
```

### DIFF
--- a/db/deploy-calibration.sh
+++ b/db/deploy-calibration.sh
@@ -6,7 +6,10 @@ cd "$(dirname "$(readlink -f "$0")")"
 # Add monorepo's node_modules to the PATH
 PATH=../node_modules/.bin:$PATH
 
-if ! wrangler d1 list | grep -q filcdn-db; then
-  wrangler d1 create filcdn-db --env calibration
+DB=filcdn-calibration-db
+ENV=calibration
+
+if ! wrangler d1 list | grep -q "$DB"; then
+  wrangler d1 create "$DB" --env "$ENV"
 fi
-wrangler d1 migrations apply filcdn-db --remote --env calibration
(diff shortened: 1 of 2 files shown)


db/deploy-calibration.sh
```diff
@@ -6,7 +6,10 @@ cd "$(dirname "$(readlink -f "$0")")"
 # Add monorepo's node_modules to the PATH
 PATH=../node_modules/.bin:$PATH
 
-if ! wrangler d1 list | grep -q filcdn-db; then
-  wrangler d1 create filcdn-db --env calibration
+DB=filcdn-calibration-db
+ENV=calibration
```
bajtos, line 10: Should I rewrite this as follows?


```suggestion
ENV=calibration
DB=filcdn-$ENV-db
```

bajtos: Done in [60861d2](https://github.com/filcdn/worker/pull/264/commits/60861d2c6e92819c9cbe1b357f01eb1097fa57d3)

pyropy: I think it's fine to keep it this way unless we also generate the database names inside `wrangler.toml` using some script to keep them in sync.

PR #273, opened 2025-09-10
Fix indexer queue

juliangruber:


indexer/wrangler.toml
```diff
@@ -38,6 +38,15 @@ database_id = "78f15bbb-391f-4797-9016-a6cb86c0b9b8"
 [env.calibration.triggers]
 crons = ["*/1 * * * *"]  # Run every minute
 
+[[env.calibration.queues.producers]]
+queue = "retry-queue"
```
bajtos, line 42: I think we need to include the environment name in the queue name, because how else would we assign different queues to different environments?

Similarly, if we want this queue to be scoped to the indexer worker only (and not e.g. the terminator worker, which uses a queue too), then I think it's best to include the worker name as well.

```suggestion
queue = "indexer-retry-queue-calibration"
```

Same comment applies to all other lines starting with `queue = `.

bajtos (changes requested):

indexer/wrangler.toml
juliangruber: I assumed queues were separated by env, but I think you're right

bajtos (approved):
👌🏻 

You may need to create these queues manually via `wrangler` CLI or the web app, IDK.

PR #282, opened 2025-09-16
add `mainnet` subgraphs

juliangruber:
I would like to figure out how to DRY this up. Asked TheChillCipher on Slack about it: https://space-meridian.slack.com/archives/C08TVNKJV7C/p1758009971861709.


subgraph/subgraph-mainnet.yaml
```diff
@@ -0,0 +1,28 @@
...
+indexerHints:
+  prune: auto
+schema:
+  file: ./schema.graphql
+dataSources:
+  - kind: ethereum
+    name: PDPVerifier
+    network: filecoin
+    source:
+      address: '0x1790d465d1FABE85b530B116f385091d52a12a3b'
```
pyropy, line 11:
```suggestion
      address: ${{ vars.PDP_VERIFIER_ADDRESS }}
```

Ditto, we could define `PDP_VERIFIER_ADDRESS` var for every environment and load it from variables.

subgraph/subgraph-mainnet.yaml
```diff
@@ -0,0 +1,28 @@
...
+      abi: PDPVerifier
+      startBlock: 5320510
```
pyropy, line 13: Ditto, this could be loaded from environment.

pyropy (changes requested):
We could probably make this DRYer by [creating deployment environments in Github Action](https://docs.github.com/en/actions/how-tos/deploy/configure-and-manage-deployments/manage-environments)s and adding variables for every environment.

bajtos (approved):
The new version looks reasonable. I don't see any obvious problems. We can address issues as they are discovered.

juliangruber:
good idea!

PR #283, opened 2025-09-17
Chore: Add FWSS mainnet address to terminator

pyropy:
This PR adds FWSS mainnet address to terminator


pyropy:
> I would continue calling it todo since we will have to update them very soon as the proper mainnet deployment is out

Makes sense!

juliangruber (approved):
nice

PR #307, opened 2025-09-29
fix: pipeline definitions

bajtos:
Tweak the definition of Goldsky pipelines to fix the problems discovered while deploying https://github.com/filbeam/worker/pull/301.

Links:
- https://github.com/filbeam/worker/issues/296

### DIFF
--- a/subgraph/pipelines/alpha-calibnet.yaml
+++ b/subgraph/pipelines/alpha-calibnet.yaml
@@ -1,5 +1,4 @@
 name: alpha-calibnet
-version: 6
 resource_size: s
 apiVersion: 3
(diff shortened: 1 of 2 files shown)


subgraph/pipelines/alpha-calibnet.yaml
```diff
@@ -1,5 +1,4 @@
 name: alpha-calibnet
-version: 6
```
bajtos, line 2: This change is not strictly needed, but I think it's better to remove the `version` field as it's out of date with the latest pipeline version on Goldsky and the CLI does not require this field to be present.

PR #312, opened 2025-09-29
Serve IPFS retrievals

bajtos:
URL schema:
- `https://1-{base32(dataSetId)}-{base32(pieceId)}.ipfs.filbeam.io/{favicon.ico}`
- `https://link.ipfs.calibration.filbeam.io/{WalletAddres}/{IpfsRootCid}/{favicon.ico}` (redirects to the above)

This is a proof of concept that I don't expect to land. Instead, I'll extract smaller pull requests and land them independently.

I will be updating `filbeam-ipfs-retriever-calibration` worker on CF with the code from this pull request.

I have a special-case path sending the retrieval requests to [our Frisbii instance](https://github.com/CheckerNetwork/frisbii-on-fly) as a workaround until Curio implements IPFS retrievals for PDP deals:

Raw image:
- https://link.ipfs.calibration.filbeam.io/0xc83dbfdf61616778537211a7e5ca2e87ec6cf0ed/bafybeig7sb5kcd3iswkv4tbrgt5hixmto3pdetxiuf723zx2i3pjlv4vae/anthony-delanoix-aDxmYZtYj7g-unsplash.jpg
- https://1-je-0.ipfs.calibration.filbeam.io/anthony-delanoix-aDxmYZtYj7g-unsplash.jpg

CAR format:
- https://link.ipfs.calibration.filbeam.io/0xc83dbfdf61616778537211a7e5ca2e87ec6cf0ed/bafybeig7sb5kcd3iswkv4tbrgt5hixmto3pdetxiuf723zx2i3pjlv4vae/anthony-delanoix-aDxmYZtYj7g-unsplash.jpg?format=car
  - DOES NOT WORK - query string is not preserved after redirect
- https://1-je-0.ipfs.calibration.filbeam.io/anthony-delanoix-aDxmYZtYj7g-unsplash.jpg?format=car

Does not work yet - we need to figure out Cloudflare config:
https://ipfs.calibration.filbeam.io/0x000000000000000000000000000000000000dead/bafybeiagrjpf2rwth5oylc64czsrz2jm7a4fgo67b2luygqjrivjbswuku/rusty-lassie.png

TODO:
- [x] Add CI & CD steps
- [x] Re-enable the skipped tests (see FIXME comments).
      This is blocked until there is at least one calibnet SP running the Curio version that scans PDP pieces for IPFS CAR blocks.
- [x] Create real SP test for subpath support (currently not supported by FOC SPs)
- [x] ~~Support all `?format=raw`~~ (later)
- [x] ~~Support for other IPFS Trustless GW options, like dag-scope, etc.~~ (later)
- [x] Calculate egress consumption, account for different size of the cache-miss CAR response vs client-side RAW response. The first iteration could report CAR size instead of the real response size. We will charge clients a bit more (CAR is larger than the raw data), but I think that's okay for a few days until we fix the problem in a follow-up pull request.
- [x] ~~Bot to peridically check IPFS retrievals~~ (later)
- [x] ~~Preserve `?format=car` and other query-string params when redirecting from `/wallet/cid` to `1-{dataset}-{piece}`.~~ (later)
- [x] Convert the wallet address to lowercase before querying the DB
- [x] Implement bad-bits lookup
- [x] Don't deploy to mainnet
- [x] Re-enable any skipped tests added in this PR

Links:
- https://github.com/filbeam/worker/issues/297
- https://github.com/filbeam/roadmap/issues/89

## TODO after merge

- [ ] Update FilBeam docs
- [ ] Update Filecoin Pin docs
- [ ] Announce the feature
- [ ] Monitor etc., then eventually also deploy to mainnet


juliangruber (changes requested):
There are quite a few things both retrievers share, like
- measure egress
- check bad bits
- etc

I think we shouldn't duplicate these. Instead, I propose:
- Create a new module `/retriever`, acting as a lib, exporting functions mentioned above and more
- Use that lib in `/ipfs-retriever`
- And `/piece-retriever`

Wdyt?

bajtos:
> There are quite a few things both retrievers share, like
> 
> * measure egress
> * check bad bits
> * etc
>   I think we shouldn't duplicate these. Instead, I propose:
> * Create a new module `/retriever`, acting as a lib, exporting functions mentioned above and more
> * Use that lib in `/ipfs-retriever`
> * And `/piece-retriever`
>   Wdyt?

That's a great idea, I like it.

My main concern is about how long it will take to ship this, I'd rather have a working PoC ipfs-retriever by Friday, even if I deploy it from an open PR.

Having said that, I am waiting for Curio to start scanning PDP Pieces for IPFS Car blocks, so I can do the suggested refactor in the meantime.

Let me give it a try and see how it goes.

bajtos:
I pushed two commits that I found on my local machine and merged the latest `main` branch. I also deployed the version https://github.com/filbeam/worker/pull/312/commits/9f2272ce9a106c1e7b3a2e4e38c5cf0f1789c6f9 to Cloudflare via `wrangler deploy --cwd ipfs-retriever --env calibration`

ipfs-retriever/lib/address.js
bajtos: I believe this file is the same as the one in the piece-retriever worker, it should be straightforward to share them via a shared package.

https://github.com/filbeam/worker/blob/c53ed018e407ca8f509824da5c7519be1363ba96/piece-retriever/lib/address.js

ipfs-retriever/lib/bad-bits-util.js
bajtos: I believe this file is the same as the one in the piece-retriever worker, it should be straightforward to share them via a shared package.

https://github.com/filbeam/worker/blob/c53ed018e407ca8f509824da5c7519be1363ba96/piece-retriever/lib/bad-bits-util.js

ipfs-retriever/lib/store.js
```diff
@@ -0,0 +1,392 @@
...
+  )
+
+  const withCDN = withPaymentRail.filter(
+    (row) => row.with_cdn && row.with_cdn === 1,
+  )
+  httpAssert(
+    withCDN.length > 0,
+    402,
+    `The Filecoin Warm Storage Service deal for payer '${payerAddress}' and ${lookupKey} has withCDN=false.`,
+  )
```
bajtos: The SQL query and some of the following filters are specific to IPFS retrievals, but many filters are shared with the Piece retrievals. It makes me wonder if it's worth extracting this into a shared helper?

ipfs-retriever/lib/store.js
```diff
@@ -0,0 +1,392 @@
...
+    // Special case for testing purposes only
+    return {
+      serviceProviderId: '9999',
+      serviceUrl: 'https://frisbii.fly.dev/',
+      dataSetId,
+      pieceId,
+      ipfsRootCid:
+        'bafybeiagrjpf2rwth5oylc64czsrz2jm7a4fgo67b2luygqjrivjbswuku',
+    }
+  }
```
bajtos: @juliangruber I believe you wanted to remove the hard-coded combination pointing to our Frisbii. I think we should remove this block as well. Thoughts?

ipfs-retriever/lib/http-assert.js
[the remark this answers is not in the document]
bajtos: See https://github.com/filbeam/worker/pull/437

ipfs-retriever/wrangler.toml
```diff
@@ -0,0 +1,59 @@
...
+# DNS_ROOT = ".ipfs.filbeam.io"
+
+# [[env.mainnet.d1_databases]]
+# binding = "DB"
+# database_name = "filcdn-mainnet-db"
+# database_id = "e8de6418-2cb7-4413-9ba0-a9c8aacf9a66"
+
+# [[env.mainnet.kv_namespaces]]
+# binding = "BAD_BITS_KV"
+# id = "7b03c39d53a041fdbe973c20285e16e9"
```
bajtos: IIRC, earlier this week, we decided to ship the IPFS retrieval worker to mainnet as well. Let's bring these mainnet wrangler config sections back.

ipfs-retriever/bin/ipfs-retriever.js
bajtos: _I want to create a discussion thread that we can mark as resolved later, and the only way to do so is by attaching the first comment to some code._

Let's review the TODO check-list in the issue description:

- Support all `?format=raw`
- Support for other IPFS Trustless GW options, like dag-scope, etc.
- Calculate egress consumption, account for different size of the cache-miss CAR response vs client-side RAW response
- Bot to peridically check IPFS retrievals
- Preserve `?format=car` and other query-string params when redirecting from `/wallet/cid to 1-{dataset}-{piece}`.

I feel many of these items can be implemented later and don't block the landing of this PR.

The only hard requirement is IMO reporting egress consumption, as it is tied to billing. I can imagine the first iteration can report CAR size instead of the real response size. We will charge clients a bit more (CAR is larger than the raw data), but I think that's okay for a few days until we fix the problem in a follow-up pull request.

It would be nice to preserve the query string when redirecting from `link.*` to `1-{dataset}-{piece}.*`, since it should be a quick fix.

I propose to convert the rest of TODOs into new GH issues and add them as children of the theme MVP of FilBeam for IPFS (https://github.com/filbeam/roadmap/issues/85).

@juliangruber thoughts?

piece-retriever/test/store.test.js
```diff
@@ -59,7 +59,7 @@ describe('getStorageProviderAndValidatePayer', () => {
 
   it('returns service provider for valid pieceCid', async () => {
     const dataSetId = 'test-set-1'
-    const pieceCid = 'test-cid-1'
+    const pieceCid = 'bafk4test'
```
bajtos: Hmm, this change in piece-retriever seems unrelated, can we revert it?

ipfs-retriever/lib/store.js
```diff
@@ -0,0 +1,392 @@
...
+
+  // We need this assertion to supress TypeScript error. The compiler is not able to infer that
+  // `withApprovedProvider.filter()` above returns only rows with `service_url` defined.
+  httpAssert(serviceUrl, 500, 'should never happen')
+
+  console.log(
+    `Validated data set ID '${dataSetId}', piece ID '${pieceId}', and service provider id '${serviceProviderId}' for ${lookupKey} and payer '${payerAddress}'. Service URL: ${serviceUrl}`,
+  )
+
+  return { serviceProviderId, serviceUrl, dataSetId, pieceId, ipfsRootCid }
```
bajtos: We will need to modify this to follow the changes made in https://github.com/filbeam/worker/pull/438

ipfs-retriever/bin/ipfs-retriever.js
```diff
@@ -0,0 +1,323 @@
...
+      // FIXME: move this logic into processIpfsResponse function
+      // When converting from CAR to RAW, set content-disposition to inline
+      // so browsers display the content instead of downloading it.
+      if (ipfsFormat !== 'car') {
+        response.headers.set('content-disposition', 'inline')
+        // Also remove the content-type header, remove x-content-type-options,
+        // and let the browser to sniff the content type.
+        response.headers.delete('content-type')
+        response.headers.delete('x-content-type-options')
+      }
```
bajtos: After we move this block to `processIpfsResponse()`, the code post-processing the response should be pretty much identical in both piece-retriever and ipfs-retriever.

ipfs-retriever/lib/store.js
```diff
@@ -0,0 +1,392 @@
...
+ *   - Performance statistics.
+ *
+ * @param {string} params.timestamp - The timestamp of the retrieval.
+ * @param {string | null} params.requestCountryCode - The country code where the
+ *   request originated from
+ * @param {string | null} params.dataSetId - The data set ID associated with the
+ *   retrieval
+ * @returns {Promise<void>} - A promise that resolves when the log is inserted.
+ */
+export async function logRetrievalResult(env, params) {
```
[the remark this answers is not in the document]
juliangruber: 9ba5aa5

bajtos:
I won't have the bandwidth to continue working on this pull request. Feel free to take my code and open a new pull request to finish the changes.

cc @juliangruber @rjan90

ipfs-retriever/bin/ipfs-retriever.js
```diff
@@ -0,0 +1,323 @@
...
+  const message =
+    isObject &&
+    status < 500 &&
+    'message' in error &&
+    typeof error.message === 'string'
+      ? error.message
+      : 'Internal Server Error'
+
+  return { status, message }
+}
```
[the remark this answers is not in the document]
juliangruber: 50ffcc2

ipfs-retriever/lib/retrieval.js
```diff
@@ -0,0 +1,232 @@
...
+ *   originEgressBytes: number | null
+ *   headers: Headers
+ * }>}
+ *   - `body` is the stream to serve to the client (raw bytes when converting from
+ *       CAR, the original body when serving CAR or passing through).
+ *       `originEgressBytes` is the number of CAR bytes read from the service
+ *       provider, or `null` when the body is passed through unchanged (in that
+ *       case the bytes served equal the bytes fetched). `headers` are the
+ *       response headers to serve, with the CAR-to-raw adjustments applied when
+ *       converting.
```
juliangruber: FIXME

ipfs-retriever/lib/retrieval.js
```diff
@@ -0,0 +1,232 @@
...
+    `${ipfsRootCid}${ipfsSubpath}`,
+    {
+      async get(blockCid) {
+        const res = await blocksReader.next()
+        if (res.done || !res.value) {
+          throw new Error(`Block ${blockCid} not found in CAR ${ipfsRootCid}`)
+        }
+        const block = res.value
+
+        // TODO: compare multihashes only
```
juliangruber: FIXME

ipfs-retriever/lib/store.js
juliangruber: https://github.com/filbeam/worker/pull/675/changes

ipfs-retriever/test/test-data-builders.js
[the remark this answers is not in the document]
juliangruber: not the right time, propose to do afterwards

juliangruber (approved):

ipfs-retriever/wrangler.toml
```diff
@@ -0,0 +1,70 @@
+name = "filbeam-ipfs-retriever"
+main = "bin/ipfs-retriever.js"
+compatibility_date = "2024-12-05"
+compatibility_flags = ["nodejs_compat"]
```
BravoNatalie: shouldn't the `compatibility_flags` have the `enable_request_signal`?  
It's on is on the `piece-retriever`

juliangruber:
TODO @juliangruber:

- [x] Create issue for updating DNS
- [x] Add @BravoNatalie to CF
- [x] Investigate: Why does the subdomain not contain wallet address?

ipfs-retriever/lib/retrieval.js
```diff
@@ -0,0 +1,220 @@
...
+  // `originEgressBytes` is final by the time we build the raw output stream.
+  let originEgressBytes = 0
+  const countingBody = (async function* () {
+    for await (const chunk of body) {
+      originEgressBytes += chunk.length
+      yield chunk
+    }
+  })()
+
+  const reader = await CarReader.fromIterable(countingBody)
```
BravoNatalie: Review the use of `fromIterable` here, from it's own docstring:

> "This performs a decode fully in memory and maintains the decoded state in memory for full access to the data via the CarReader API. Care should be taken for large archives; this API may not be appropriate where memory is a concern or the archive is potentially larger than the amount of memory that the runtime can handle."

Consider using `CarBlockIterator`:

>  "Rather than decoding the entire byte array prior to returning the iterator, as in `CarReader.fromIterable`, only the header is decoded and the remainder of the CAR is parsed as the Blocks as yielded."

```ts
// Streaming — only header decoded upfront, blocks yielded lazily
const iterator = await CarBlockIterator.fromIterable(asyncIterable)
const blocksReader = iterator[Symbol.asyncIterator]()
// blocksReader.next() — same call pattern as the current code
```

retrieval/lib/access.js
```diff
@@ -0,0 +1,161 @@
...
+    'data_set_egress_quotas.cdn_egress_quota',
+    'data_set_egress_quotas.cache_miss_egress_quota',
+    'service_providers.service_url',
+    'service_providers.is_deleted AS service_provider_is_deleted',
+    'wallet_details.is_sanctioned',
+    ...extraColumns,
+  ]
+
+  return `
+    SELECT ${columns.join(', ')}
```
BravoNatalie, line 149: In the case `WHERE pieces.ipfs_root_cid = ?` (ipfs-retriever/lib/store.js:14) this query will present problems at scale.

With Storacha's sharding model, one `ipfsRootCid` maps to N Filecoin pieces (shards), all stored on the same SP. Diferent from the filecoin-pin, where one upload is equivalent to 1 piece and 1 `ipfsRootCid`. 

The query returns one row per piece, no DISTINCT, no GROUP BY, so `validateQueryResultsAndGetCandidates` produces N candidates all with the same `serviceUrl`.

`selectRetrievalCandidate` then retries randomly through those N candidates on failure. But `retrieveIpfsContent` only uses `serviceUrl` and `ipfsRootCid`, `pieceId` is never passed to it. So each "retry" makes the exact same HTTP request to the same SP. A genuinely failing SP gets hit N times before the 502.

db/migrations/0028_retrieval_logs_cache_miss_egress_bytes.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE retrieval_logs ADD COLUMN cache_miss_egress_bytes INTEGER;
```
BravoNatalie, line 1: Missing index on `pieces.ipfs_root_cid`

The only index on the `pieces` table is `pieces_cid`. `ipfs_root_cid` has no index, so the query `buildRetrievalCandidateQuery` with `WHERE pieces.ipfs_root_cid = ?` does a full table scan on every `ipfs-retriever` request. With millions of pieces this degrades linearly.

.github/workflows/ci.yml
```diff
@@ -80,6 +80,14 @@ jobs:
           accountId: ${{ secrets.CLOUDFLARE_ACCOUNT_ID }}
           preCommands: ../db/deploy-${{ matrix.environment }}.sh
           environment: ${{ matrix.environment }}
+      - name: Deploy IPFS Retriever and Migrate Database
+        uses: cloudflare/wrangler-action@v3
```
BravoNatalie: use `cloudflare/wrangler-action@v4`? to match the other workers

ipfs-retriever/wrangler.toml
```diff
@@ -0,0 +1,70 @@
...
+binding = "BAD_BITS_KV"
+id = "2f2e5486ea0c48e993f6dff87a4aa102"
+
+[vars]
+ORIGIN_CACHE_TTL = 86400
+CLIENT_CACHE_TTL = 31536000
+DNS_ROOT = ".localhost"
+BOT_TOKENS = ""
+ENFORCE_EGRESS_QUOTA = false
+
```
BravoNatalie, line 23: Shouldn't this worker also have the `tail_consumers` configs?

ipfs-retriever/wrangler.toml
juliangruber: 04dbafeb804b3a3cb882ced81bf1c695e9ac92b5

ipfs-retriever/wrangler.toml
juliangruber: 04dbafeb804b3a3cb882ced81bf1c695e9ac92b5

.github/workflows/ci.yml
juliangruber: da457d9ca940270dfa71082cdc4f6a893156a274

db/migrations/0028_retrieval_logs_cache_miss_egress_bytes.sql
juliangruber: 917089945c9a42dbbdd8387f95ad550079ddf56f

.github/workflows/ci.yml
```diff
@@ -80,6 +80,14 @@ jobs:
...
+        with:
+          workingDirectory: ipfs-retriever
+          apiToken: ${{ secrets.CLOUDFLARE_API_TOKEN }}
+          accountId: ${{ secrets.CLOUDFLARE_ACCOUNT_ID }}
+          preCommands: ../db/deploy-${{ matrix.environment }}.sh
+          environment: ${{ matrix.environment }}
```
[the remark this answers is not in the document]
juliangruber: d0c3fe78c6f2825b29fab6ad21be19c5167387e8

ipfs-retriever/package.json
```diff
@@ -0,0 +1,23 @@
+{
+  "name": "@filbeam/ipfs-retriever",
+  "version": "1.0.0",
+  "private": true,
+  "description": "FilBeam IPFS Retrieval Worker",
+  "author": "Space Meridian <filbeam@meridian.space>",
+  "type": "module",
+  "main": "bin/indexer.js",
+  "scripts": {
```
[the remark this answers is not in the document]
juliangruber: b2d5f0f1fc4f8a2c217a5a6e9b8d917dc38483e8

ipfs-retriever/package.json
```diff
@@ -0,0 +1,23 @@
...
+    "build:types": "wrangler types",
+    "deploy:calibration": "wrangler deploy --env calibration",
+    "deploy:mainnet": "wrangler deploy --env mainnet",
+    "start": "wrangler d1 migrations apply dev-db --local --env dev --cwd ../db && wrangler dev --env dev",
+    "test": "wrangler d1 migrations apply test-db --local --cwd ../db && vitest run"
+  },
```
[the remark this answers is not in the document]
juliangruber: b2d5f0f1fc4f8a2c217a5a6e9b8d917dc38483e8

ipfs-retriever/test/test-data.js
```diff
@@ -0,0 +1,30 @@
+/**
+ * @type {{
+ *   serviceProviderId: string
+ *   serviceUrl: string
+ *   ipfsRootCid: string
+ *   dataSetId: number
+ *   pieceId: string
+ * }[]}
+ */
```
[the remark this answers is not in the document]
juliangruber: 55a0839ebd284af928273fee20a9437baccc3a3b

ipfs-retriever/wrangler.toml
```diff
@@ -0,0 +1,83 @@
...
+[env.calibration]
+tail_consumers = [{ service = "filbeam-tail-handler-calibration" }]
+
+[env.calibration.vars]
+ENVIRONMENT = "calibration "
+ORIGIN_CACHE_TTL = 86400
+CLIENT_CACHE_TTL = 31536000
+DNS_ROOT = ".ipfs.calibration.filbeam.io"
+ENFORCE_EGRESS_QUOTA = true
```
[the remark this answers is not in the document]
juliangruber: 63253e99bcc9a2dc5d214ec540585244a0f8a9bb

retrieval/lib/bot-auth.js
```diff
@@ -0,0 +1,35 @@
...
+ * Resolves the bot name for a request's Bearer token, or `undefined` for
+ * anonymous requests. Throws a 401 for a malformed Authorization header or an
+ * unknown token.
+ *
+ * @param {Request} request
+ * @param {object} args
+ * @param {string} args.BOT_TOKENS - JSON object mapping access token to bot
+ *   name.
+ * @returns {string | undefined} Bot name or the access token
+ */
```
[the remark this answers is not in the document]
juliangruber: 5724d36f8ff0c0576fdde2c68a54bbac0e4e08be

retrieval/lib/bot-auth.js
```diff
@@ -0,0 +1,35 @@
...
+export function checkBotAuthorization(request, { BOT_TOKENS }) {
+  const botTokens = JSON.parse(BOT_TOKENS)
+
+  const auth = request.headers.get('authorization')
+  if (!auth) return undefined
+
+  const [prefix, token, ...rest] = auth.split(' ')
```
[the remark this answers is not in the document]
juliangruber: 5724d36f8ff0c0576fdde2c68a54bbac0e4e08be

ipfs-retriever/lib/retrieval.js
```diff
@@ -0,0 +1,220 @@
...
+  const blocksReader = reader.blocks()
+
+  const entries = exporter(
+    `${ipfsRootCid}${ipfsSubpath}`,
+    {
+      async get(blockCid) {
+        const res = await blocksReader.next()
+        if (res.done || !res.value) {
+          throw new Error(`Block ${blockCid} not found in CAR ${ipfsRootCid}`)
+        }
```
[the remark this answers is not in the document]
juliangruber: Good catch in the general case. Leaving this as a follow-up rather than fixing here:

- The calibration content this POC serves is single-block (raw), and trustless-gateway CARs come back `order=dfs`, which matches the exporter's traversal, so the sequential reader works for what we serve today.
- A correct order-independent fix needs to preserve the existing multihash-equivalence handling (a block stored under a different codec/CID version with the same multihash still matches), so it's a multihash-keyed index rather than a plain `reader.get(blockCid)`, plus a multi-block out-of-order test to lock it in.

Tracking it as a follow-up for multi-block support.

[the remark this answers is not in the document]
juliangruber: cc @BravoNatalie

retrieval/lib/access.js
juliangruber: Great catch! https://github.com/filbeam/worker/pull/689 for your consideration

ipfs-retriever/lib/retrieval.js
juliangruber: 24d6691d2020f0bccc34751cd8c1dab37b56f228

usage-reporter/test/test-helpers.js
```diff
@@ -97,6 +97,8 @@ export const randomId = () => String(Math.ceil(Math.random() * 1e10))
  * @param {string} params.dataSetId - Data set ID
  * @param {number} params.responseStatus - HTTP response status (default: 200)
  * @param {number | null} params.egressBytes - Egress bytes (default: null)
+ * @param {number | null} params.cacheMissEgressBytes - CAR bytes fetched from
+ *   the SP on a cache miss (default: null)
  * @param {number} params.cacheMiss - Cache miss flag (0 or 1, default: 0)
  * @param {boolena} params.cacheMissResponseValid
  */
```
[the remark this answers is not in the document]
juliangruber: 325b693f168fb1db78a139f749df4ea6e8baca58

juliangruber:
@BravoNatalie no need to update DNS records post merge, they are already pointing at the right worker. The merge will just redeploy that worker

juliangruber:
>  Investigate: Why does the subdomain not contain wallet address?

data set id is enough to look up the wallet address. The wallet address isn't used, as it's too big for subdomains. This makes subdomains possible.

In order to still be able to request the piece from multiple SPs (for redundancy), the worker looks up the root CID for given data set & piece, and considers all SPs having that piece.

juliangruber:
Manual deployment of IPFS retrievals on FilBeam still works:


- https://link.ipfs.calibration.filbeam.io/0xaEF30c4730CF03b5dE26b57007A0e4a9752df82b/bafkreiheygfzn22dfeos3xoay5cxnfb464znd2rszieyzcinlsgu2z7kau/
- https://1-hdyq-0.ipfs.calibration.filbeam.io/

ipfs-retriever/lib/retrieval.js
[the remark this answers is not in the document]
BravoNatalie: issue #692

docs/ipfs-retriever.md
```diff
@@ -0,0 +1,185 @@
...
+---
+
+## URL formats
+
+There are two entry points:
+
+### Slug subdomain (primary)
+
+```
+https://1-{base32(dataSetId)}-{base32(pieceId)}.ipfs.calibration.filbeam.io/{subpath}
```
rvagg, line 18: how do we derive ipfsRootCid from this path? do we keep a local database of data set metadata for every data set with CDN turned on so we have the mapping locally?

rvagg: section below clarifying this if it's as simple as "we have the data", cause it's the same kind of question as the wallet-in-subdomain one

docs/ipfs-retriever.md
```diff
@@ -0,0 +1,185 @@
...
+The initial design proposed combining the IPFS CID and wallet address into a single subdomain component (e.g. `bafk123-0xabc.filbeam.io`). This was ruled out because a single DNS label is limited to 63 characters, too short to fit both a base58 CID and a 42-character Ethereum address together. (see [follow-up comment, issue #297](https://github.com/filbeam/worker/issues/297#issuecomment-3352714474))
+
+#### Why the wallet address is not in the slug at all
+
+The wallet address is not needed in the slug because `dataSetId` alone is sufficient to look it up in D1. Omitting it keeps the subdomain short enough to be valid. The worker resolves the wallet from the dataset record at query time. (see [review comment, PR #312](https://github.com/filbeam/worker/pull/312#issuecomment-4797500253))
+
+### Bare domain redirect (convenience)
+
+```
+https://ipfs.calibration.filbeam.io/{walletAddress}/{ipfsRootCid}/{subpath}
```
rvagg, line 49: does this path validate that this walletAddress is actually associated with that ipfsRootCid?

docs/ipfs-retriever.md
```diff
@@ -0,0 +1,185 @@
...
+```
+
+Handled by `handleDnsRootRequest` in `ipfs-retriever/bin/ipfs-retriever.js`. Looks up the slug for the given wallet and CID in D1, then issues a 302 redirect to the slug subdomain URL. Useful for constructing shareable links without knowing the on-chain IDs upfront.
+
+If no path is provided (`/`), redirects to `https://filbeam.com`.
```
rvagg: no path? as in `https://ipfs.calibration.filbeam.io/` redirects? or no "subpath"? some clarification on this line would be good

docs/ipfs-retriever.md
```diff
@@ -0,0 +1,185 @@
...
+
+### Query parameters
+
+- `?format=car` — skip CAR-to-raw conversion and serve the raw CAR archive to the client
+- `?format=raw` — not yet implemented (returns 400); tracked in [issue #295](https://github.com/filbeam/worker/issues/295)
```
rvagg: What would this return? just the terminal block for the requested subpath? what if the terminal is a many-block file (such as a video file)? the reason this is fine for public ipfs gateways is because you can hit it with arbitrary CIDs, but we're limiting to strictly entering via the on-chain recorded ipfsRootCid plus a subpath, but there's lots of complexity in what that can  take you to. Even a directory listing can be many hundreds of blocks to represent (the wikipedia directory under /en/ is thousands of blocks for example); I think maybe this should be descoped until a need arises.

docs/ipfs-retriever.md
```diff
@@ -0,0 +1,185 @@
...
+https://ipfs.calibration.filbeam.io/{walletAddress}/{ipfsRootCid}/{subpath}
+```
+
+Handled by `handleDnsRootRequest` in `ipfs-retriever/bin/ipfs-retriever.js`. Looks up the slug for the given wallet and CID in D1, then issues a 302 redirect to the slug subdomain URL. Useful for constructing shareable links without knowing the on-chain IDs upfront.
+
+If no path is provided (`/`), redirects to `https://filbeam.com`.
+
+### Query parameters
+
+- `?format=car` — skip CAR-to-raw conversion and serve the raw CAR archive to the client
```
rvagg: but we also skip validation of the CAR? so we pass it on to the client and it becomes their problem to validate? might be fine, but if this is part of the service we're advertising then we should be sure that they follow standard trustless-http protocol for validation of the contents of the CAR.

docs/ipfs-retriever.md
```diff
@@ -0,0 +1,185 @@
...
+- No `?format` — default; converts CAR to raw file bytes
+
+---
+
+## Request flow
+
+```
+1. handleFetchRequest()       — shared fetch lifecycle (auth, error handling, egress logging)
+2. parseRequest()             — decode slug → dataSetId + pieceId + subpath + format
+3. getRetrievalCandidatesByDataSetAndPiece() — query D1 for SP candidates
```
rvagg: this step looks like it's hiding a bunch of stuff that someone more familiar with filcdn would know, but me, as someone unfamiliar, has no idea about! does this get you just a list of SPs? does it get you the ipfsRootCid(s) for the dataSetId+pieceId combo? And why would there be more than 1 sp candidate for this? Are we cross-referencing other providers this wallet stored the same piece with? If we do that do we also cross reference that the metadata makes the same claims about the relationship with ipfsRootCid? 🤯 so many questions I have about this one step! maybe there's a doc for this already.

docs/ipfs-retriever.md
```diff
@@ -0,0 +1,185 @@
...
+
+### Why `blockReadConcurrency: 1`
+
+The exporter is called with `{ blockReadConcurrency: 1 }`. This forces it to request blocks strictly one at a time in DFS traversal order. The blockstore's `get(cid)` does not look blocks up by CID, it calls `blocksReader.next()` and expects that the next block in the CAR is always the one being requested. This works because the SP guarantees DFS-ordered delivery (see [SP integration contract](#sp-integration-contract)). If `blockReadConcurrency` were greater than 1, the exporter would request blocks in parallel and the sequential CAR reader would return the wrong block for each.
+
+### `CarBlockIterator` vs `CarReader`
+
+`CarBlockIterator.fromIterable` reads only the CAR header (roots + version) upfront. Block data is pulled lazily as the iterator is consumed.
+
+### CAR-to-raw conversion
```
rvagg, line 197: you have directory entries below, but what about the two cases for non-directories, where you hit file contents but may not know what to do with it; you kind of alude to it in this section but let's be more clear:

1. The root CID itself resolves to some raw content, you don't actually know what it is, you don't even have a filename (we did away with directory wrapping for filecoin-pin in https://github.com/filecoin-project/filecoin-pin/pull/448 so it'll be common even for our content). I think this section says that we just shove it out to the browser and make it their problem to figure out the content type by magic bytes sniffing, I guess that works with modern browsers now.
2. We have a filename, so at least have an extension for the file and _could_ give a reasonable content-type for it like a standard web server. Do we not even do that here and just send it out and make it the browser's problem? Do you know if this is fine and good for modern browsers to render a whole website, JS, CSS, HTML, images, videos, etc.? Is there any reason this might be a risky strategy?

Either way, let's just note that there are two paths here -- we retrieve content that we just know are raw bytes and know nothing else about it, and the case where we at least have the terminal path that tells us the filename which may have an extension.

docs/ipfs-retriever.md
```diff
@@ -0,0 +1,185 @@
...
+`processIpfsResponse` uses the `recursive` export from `ipfs-unixfs-exporter` aliased as `exporter`. Despite the name, it is used here only to resolve the first entry (the requested path) and stream its bytes, the loop exits after the first iteration intentionally (`// eslint-disable-next-line no-unreachable-loop`). The actual `exporter()` function would be semantically cleaner for this use case.
+
+When converting CAR to raw:
+
+- `content-disposition: inline` is set so browsers display the content instead of downloading it
+- `content-type` and `x-content-type-options` are removed so the browser sniffs the raw bytes
+
+### Directory entries
+
+If the resolved path is a UnixFS directory, the worker returns **404 Not Found** (`retrieval.js:195`). Directory listing is not implemented. Since the SP returns a path-scoped CAR with `dag-scope=all`, the directory block and immediate child blocks are present in the CAR. The 404 is a choice, not an architectural limit. This is tracked in [issue #696](https://github.com/filbeam/worker/issues/696).
```
rvagg, line 209: see my note there about dag-scope=entity, I think it'd be a better choice in general for everything we're doing here than all.

docs/ipfs-retriever.md
```diff
@@ -0,0 +1,185 @@
...
+
+## SP integration contract
+
+The pipeline makes specific assumptions about how the SP delivers the CAR. These are satisfied by **Curio**, which implements the [frisbii](https://github.com/ipld/frisbii) trustless HTTP gateway:
+
+1. **Path-scoped CAR** — the SP is called as `GET /ipfs/{rootCid}{subpath}?format=car`. It returns a CAR containing exactly the blocks needed to walk from `rootCid` to `subpath` and read the file. No more, no less. There are no wasted bytes for single-asset requests regardless of dataset size.
+
+2. **DFS-ordered blocks** — blocks are written to the CAR in the same order the DAG traversal engine requests them (depth-first). This is enforced in frisbii via `carPipe`, which hooks into the IPLD link system and writes each block to the CAR immediately as it is loaded during traversal. This is what makes the sequential `blocksReader.next()` blockstore safe.
+
+3. **Complete or fail loudly** — if any block is missing from SP storage, the traversal fails before any bytes are written to the CAR. The SP returns an HTTP error, not a partial CAR. The worker either gets a complete, valid CAR or an error response, never a silently truncated one.
```
rvagg: Well ... not quite this clean. we made a choice in the frisbii version of trustless http that we'd prioritise streaming, so while traversing, it's possible to find that we don't have a block that we need, perhaps the client didn't upload their full DAG, and we have no useful way of signalling an error in that case, trailers are not a useful error signalling mechanism and we decided not to build in some kind of error signal into the CAR we returned, so we must rely on the fact that the receiver is doing the identical traversal to also discover that the _next_ block isn't present and the CAR terminated prematurely. So it is possible start getting data, be happily traversing down the DAG with unixfs-exporter and then just halt. In which case we need to return some kind of meaningful error. It _could_ be an SP problem, maybe their system is busted in some way, maybe network got disconnected and we don't get a clean disconnect error, but it could also be that the user didn't actually store the entire DAG with that SP.

The nice thing about IPFS retrievals and how they work on the SP side is that the entire DAG doesn't need to be in the same piece. You could have one tiny piece that just has the root CID and then a hundred more pieces that contain CARs with all the rest of  the blocks for your content, and frisbii retrieval will just fetch the blocks in the right order out of all of those pieces to serve back to you (so the 1Gb piece limit we currently have in place doesn't stop you from storing and serving a 100Gb movie file using IPFS retrievals). **BUT** unlike Rainbow and other IPFS gateway software that taps into a broader network, using the DHT or IPNI to discover providers for blocks, the DAG _must_ be all on the same SP for any of this to work. You can't shard your DAG across multiple SPs to have the filbeam retrieval work; but you _can_ shard it and have generic IPFS gateway retrievals work because it does block-by-block lookups as it traverses and can skip from SP to SP to find what it needs. Important distinction that we should probably note somewhere.

docs/ipfs-retriever.md
```diff
@@ -0,0 +1,185 @@
...
+
+**If a non-frisbii SP is ever onboarded**, all three guarantees must be verified before integration (tracked in [issue #692](https://github.com/filbeam/worker/issues/692)). The CID mismatch error (`Unexpected block CID`) is the failure mode if ordering is violated, it is not obvious without this context.
+
+### Alternative: block-by-block retrieval (Discarded)
+
+An alternative design (used by IPFS Shipyard tooling like Boxo and verified-fetch) drives the exporter with a blockstore whose `get(cid)` makes individual HTTP requests: `GET /ipfs/{blockCid}?format=raw`. This eliminates the ordering dependency and enables per-block Cloudflare caching.
+
+The trade-off: a single file goes from 1 SP fetch to potentially 20–80 individual fetches. For FilBeam, which controls both ends and knows all blocks are at the same SP, the CAR approach is currently a good solution, one round-trip, any file size, fully streaming. The block-by-block design exists to solve distributed-network problems (blocks scattered across unknown peers) that FilBeam does not have.
```
rvagg: On the other side of the equation here is that there's caching potential for block-by-block, that should be noted, maybe it's in the slack thread linked; /path/to/some/deep/nested/content.html where that html includes 100 jpgs has to re-fetch the DAG for /path/to/some/deep/nested/ for each jpg. Typically these are quite small and it's not really a big deal. There are some cases where it might matter but on balance I'd say that for our narrow case where the client is choosing  to opt in to this service by storing the entire DAG on one of the SPs in this service rahter than scattering it around the internet and crossing their fingers, CAR retrieval makes more sense. But it's not a clear-cut case I think.

docs/ipfs-retriever.md
```diff
@@ -0,0 +1,185 @@
...
+}
+```
+
+- `cacheEverything: true` — caches the SP response regardless of its `Cache-Control` header
+- 2xx responses are cached for `ORIGIN_CACHE_TTL` seconds (currently **86400, 1 day**)
+- 404 and 5xx responses are never cached
+
+**Cache key:** `{spBaseUrl}/ipfs/{ipfsRootCid}{ipfsSubpath}?format=car`
+
+Since CIDs are content-addressed and immutable, the same key always resolves to the same bytes. `ORIGIN_CACHE_TTL` could theoretically be much longer, but a very large TTL risks filling PoP cache storage with large CARs. The current 1-day value is a reasonable starting point and should be revisited once cache hit rates and storage pressure are measurable.
```
rvagg: also reinforced by my point above that perhaps the SP is having a bad day and their database had problems resolving all the blocks in the CAR, or one of their disks holding a piece that contained part of the DAG was down for a moment and you got an incomplete CAR;  you don't want to associated strictly a forever-cache with a potential blip in serving.

It's true that CIDs point to immutable content, but you're dealing with a whole string of CIDs that must be validated together. In theory CID+subpath is unique, but there a ways for it to go bad and not know until you run the traversal.

docs/ipfs-retriever.md
```diff
@@ -0,0 +1,185 @@
...
+
+**Cache miss detection:** after the fetch, `CF-Cache-Status: HIT` means the edge served a cached copy; anything else is treated as a cache miss. Cache misses are what drive egress quota billing.
+
+### Layer 2 — Client/browser cache
+
+Set on every successful response to the client (`retrieval/lib/fetch-handler.js`):
+
+```
+Cache-Control: public, max-age=31536000
+```
```
rvagg, line 288: You do have the same problem as I outlined above I think if we're streaming the content to the user unless you hard-abort the connection or something (IIRC there's no good, reliable signal for terminating an HTTP stream that would invalidate a cache? I think I remember we wrestled with this for a while in the trustless HTTP design). Might be worth a bit of research to make sure that this isn't going to cause problems -- tell the client that their 100Mb movie file is correct for a year, only able to fetch and serve 50Mb of it, what then? is the movie now 50Mb for that client for a year? Can we terminate in a way that will reliably tell the client that the max-age shouldn't be respected?

rvagg: if in doubt, we should try and pull in `@lidel` to get some answers since this kind of nuance is right up his alley

docs/ipfs-retriever.md
bajtos: Related: https://space-meridian.github.io/docs/Engineering/fefc0d538c414c8e96a56587e4ca75ce/FilBeam/FilBeam%20URL%20format%2027ecdd5cccdb806caeeaefad80cbf64d.html#27ecdd5c-ccdb-800e-ac46-de8a4a64f0a8

> how do we derive ipfsRootCid from this path? do we keep a local database of data set metadata for every data set with CDN turned on so we have the mapping locally?

Yes, see https://github.com/filbeam/worker/pull/310

docs/ipfs-retriever.md
BravoNatalie: yes, but only rely on indexed data

docs/ipfs-retriever.md
BravoNatalie: Yes, we don't validate it. 
Isn't that the expected behavior according to the Trustless Gateway protocol?

I've made that explicit in the documentation.
https://github.com/filbeam/worker/pull/312/changes#diff-cfd04945276d0a793adc966c6641b4a4c93606dc159a46e3ad3195ec757a096aR58

docs/ipfs-retriever.md
BravoNatalie: I updated it with more details

BigLep:
@BravoNatalie : I'm going to put this back to TODO since I know this work is currently on pause. If there are still "TODOs after merge" that are listed in the issue description, lets capture those in the project issue https://github.com/filbeam/worker/issues/297 which will stay open after the merge so we don't lose track of them.

PR #315, opened 2025-09-30
Update use explicit cf caching. Closes #313

### LINKED ISSUE #313: Use explicit CF caching
In https://github.com/bajtos/cf-latencies?tab=readme-ov-file @bajtos realized that there's a notable performance increase from

explicit caching:

https://github.com/bajtos/cf-latencies/blob/85cde15d3d004c343450d0cf3e52cbc31f1901b3/src/index.ts#L126

over default caching:

https://github.com/bajtos/cf-latencies/blob/85cde15d3d004c343450d0cf3e52cbc31f1901b3/src/index.ts#L99-L104

juliangruber:
Closes #313

TS doesn't like the use of the global `caches`, which does exist in CF. How do we fix this?


retriever/lib/retrieval.js
```diff
@@ -18,27 +18,32 @@
...
-    },
-    signal,
-  })
-  const cacheStatus = response.headers.get('CF-Cache-Status')
-  if (!cacheStatus) {
-    console.log(`CF-Cache-Status was not provided for ${url}`)
-  }
 
-  const cacheMiss = cacheStatus !== 'HIT'
+  const cacheKey = new Request(url, request)
```
bajtos, line 27: Since the `fetch` call is not using any additional information from the original `request`, I think we shouldn't be using the original request for constructing the cache key. In particular, we should ignore request headers provided by the FilBeam client.

Would the following work?

```js
const originReq = new Request(url)
const cacheLKey = originReq
// ...
const response = await fetch(originReq)
```

retriever/lib/retrieval.js
```diff
@@ -18,27 +18,32 @@
...
+  let response = await caches.default.match(cacheKey)
```
pyropy, line 28: Did you try adding [@cloudflare/worker-types](https://www.npmjs.com/package/@cloudflare/workers-types?activeTab=readme) package? Maybe that could resolve this issue.

bajtos (commented):
The implementation LGTM, but we need to find a way how to fix the TypeScript error.

Before we land this PR, I would like us to have tooling to observe the impact of this change on the retrieval response time, as discussed [in Slack](https://space-meridian.slack.com/archives/C099XQTBGLU/p1759733616199559).

retriever/lib/retrieval.js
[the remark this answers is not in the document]
juliangruber: Updating the compatibility date unfortunately doesn't fix TS

PR #323, opened 2025-10-01
Add KV cache for D1 index. Closes #306

### LINKED ISSUE #306: Use KV for index lookup
Querying a D1 SQL for every index lookup during retrieval request is too slow. Move or copy the relevant parts of the index to a KV store, with O(1) lookup complexity.

juliangruber:
~~As in https://github.com/filbeam/worker/pull/315, TS errors that I don't know how to fix are making CI fail.~~

Closes #306


pyropy (commented):
Changes look good. I'd only like to discuss cache design before giving it a approval.

piece-retriever/bin/piece-retriever.js
```diff
@@ -83,23 +84,41 @@ export default {
...
+          `Unsupported Service Provider: ${serviceProviderId}`,
+        )
 
-      httpAssert(
-        serviceProviderId,
-        404,
-        `Unsupported Service Provider: ${serviceProviderId}`,
-      )
+        ctx.waitUntil(
+          env.KV.put(indexCacheKey, JSON.stringify([dataSetId, serviceUrl])),
```
bajtos, line 119: What is the TTL of these KV entries?

The badbits check is valid for 5 minutes, then we need to re-check.

https://github.com/filbeam/worker/blob/d5f8523642d5349483759874629350377d2a9d24/bad-bits/wrangler.toml#L44

The mapping from PieceID to DataSetId + ServiceURL is valid until a piece or dataset is deleted. When that happens, your new code will clear the KV entries, so all is good there 👍🏻

So we need to discuss bad-bits checks. Two options come to my mind:
1. Rework the bad-bits worker to maintain the bad-bits set in KV (either only in KV or in addition to SQL). The tricky part is how to detect deletions. Then the retriever worker can run two KV queries - one for the cached lookup of dataset+provider, the other for bad-bits check.
2. Change the bad-bits worker to invalidate all cached KV entries where the key contains a bad-bits CID. The problem is that the keys include a wallet address, so I think we would have to enumerate all keys.

bajtos (changes requested):
Great start!

indexer/lib/fwss-handlers.js
```diff
@@ -78,4 +78,27 @@ export async function handleFWSSServiceTerminated(env, payload) {
   )
     .bind(String(payload.data_set_id))
     .run()
+  await clearDataSetIndexCache(env, payload.data_set_id)
+}
+
+/**
+ * @param {Env} env
```
bajtos, line 85: You can use the approach we use elsewhere.

```
 * @param {{ CHAINALYSIS_API_KEY: string; DB: D1Database, KV: KVNamespace }} env
 ```

File: piece-retriever/bin/piece-retriever.js (hunk 1 above)
Comment (bajtos) [reply piece-retriever/bin/piece-retriever.js:119]: Option 3 that came to my mind - I think that may be the simplest solution:
- Cache bad-bits lookup under a different key, e.g. `bad-bits/${cid}`. 
- Include the time of the check in the value, e.g. `{ isBadBit: true/false, checkedAt: timestamp }`
- In the retrieval worker, if `checkedAt` is more than 5 minutes ago, then query the bad-bits DB and update the cached KV entry.

Anyhow, feel free to come up with a different design!

File: indexer/lib/pdp-verifier-handlers.js [hunk 3]
```diff
@@ -44,3 +45,29 @@ export async function removeDataSetPieces(env, dataSetId, pieceIds) {
...
+async function clearDataSetPiecesIndexCache(env, dataSetId, pieceIds) {
+  const { results } = await env.DB.prepare(
+    `
+      SELECT data_sets.payer_address AS payerAddress, pieces.cid AS pieceCID
+      FROM data_sets
+      INNER JOIN pieces ON pieces.data_set_id = data_sets.id
+      WHERE data_sets.id = ? AND pieces.id IN (${new Array(pieceIds.length)
+        .fill(null)
+        .map(() => '?')
+        .join(', ')})
```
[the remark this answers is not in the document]
juliangruber: What do you think about fe5d246? This is the only file where the same pattern was used

bajtos (changes requested):
Please re-request another review after you implement the change we agreed on yesterday, where the bad-bits worker will use the KV store only, no D1 database.

bad-bits/lib/store.js
```diff
@@ -1,58 +1,106 @@
...
+    i++
+  ) {
+    await env.KV.put(
+      `bad-bits:_latest-hashes:${i}`,
+      [...latestHashes]
+        .slice(
+          i * KV_SEGMENT_MAX_HASH_COUNT,
+          (i + 1) * KV_SEGMENT_MAX_HASH_COUNT - 1,
+        )
+        .join(','),
```
bajtos: Please move this code to `utils.js` alongside `getAllBadBitHashes`, to make it easy to reason about both read and write paths.

Also, there is a bug here: if the number of bad-bit entries shrinks, and we can store all entries in less buckets than before, then your code will not delete those old buckets (`bad-bits:_latest-hashes:${i}` keys) that should not be used any more.

Suggestion how to fix this:
- After writing all buckets, create an empty bucket with the key `bad-bits:_latest-hashes:${N}`, where `N = Math.ceil(latestHashes.size / KV_SEGMENT_MAX_HASH_COUNT)`.
- In `getAllBadBitHashes()`, stop iterating over the buckets after you find an empty bucket.

bajtos (changes requested):

bad-bits/test/bad-bits.test.js
```diff
@@ -103,11 +102,6 @@ describe('fetchAndStoreBadBits', () => {
       'If-None-Match in the first request',
     ).toBe(null)
 
-    // The history is updated when new bad bits are fetched
-    expect(await getBadBitsHistory(env)).toStrictEqual([
-      { timestamp: lastUpdateAt.toISOString(), etag },
-    ])
```
bajtos: I guess we don't need to keep an audit trail of all bad-bit updates we performed. However, I think we should at least keep the timestamp of when we finished the last full update. To show that we are actively keeping up with bad-bits updates.

bad-bits/lib/store.js
```diff
@@ -1,58 +1,106 @@
...
     )
   }
+  if (etag) {
+    await env.KV.put('bad-bits:_latest-etag', etag)
+  }
```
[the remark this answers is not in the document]
juliangruber: https://github.com/filbeam/worker/pull/367/files#r2445059959

bajtos (approved):
I love how much simpler this pull request became after we removed the changes related to bad-bits 👏🏻

indexer/lib/fwss-handlers.js
```diff
@@ -78,4 +78,27 @@ export async function handleFWSSServiceTerminated(env, payload) {
...
+      INNER JOIN pieces ON pieces.data_set_id = data_sets.id
+      WHERE data_sets.id = ?
+    `,
+  )
+    .bind(String(dataSetId))
+    .run()
+  await Promise.all(
+    results.map(async ({ payerAddress, pieceCID }) => {
+      await env.INDEX_CACHE_KV.delete(`${payerAddress}/${pieceCID}`)
+    }),
```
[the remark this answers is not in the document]
bajtos: Yes, on second thought, I also concluded that the limit of 1000 pieces per dataset is too low, and we need to explore other options.

Considering the complexities, maybe we should put this performance optimisation on hold until the GA launch. WDYT?

PR #324, opened 2025-10-01
feat: add payment settlement worker 

pyropy:
### Description

This PR introduces a worker that will periodically settle all payment rails for data sets that have or had CDN service enabled. 

### Payment Rail Settlement

The worker will periodically query data sets that meet the following criteria: usage was reported within the past 30 days, funds can still be settled, and the data set has not been terminated due to sanctiones. It will then attempt to execute payment rail settlements for these data sets.

### Terminated Rail Settlement

Upon a service termination indexer will set  `lockup_unlocks_at` to a data set that's calculated as `epoch + DEFAULT_LOCKUP_PERIOD`. During that period (prior to `lockup_unlocks_at`) settlement worker will be able to settle payments for egress that has occurred prior to termination but was reported after. 

### Changelog

- Add new rail-settlement worker to handle payment rail settlements
- Add database migration for `lockup_unlocks_at` tracking
- Update indexer to track settlement epochs via chain module
- Monitor settlement transactions and retry on failure
- Configure worker with scheduled cron for daily settlements
- Add tests for settlement logic

### Dependencies

- https://github.com/filbeam/contracts/pull/1

Closes https://github.com/filbeam/roadmap/issues/87 
Partially closes https://github.com/filbeam/roadmap/issues/83

### DIFF
--- a/payment-settler/lib/chain.js
+++ b/payment-settler/lib/chain.js
@@ -2,13 +2,7 @@ import { createPublicClient, createWalletClient, http } from 'viem'
 import { privateKeyToAccount } from 'viem/accounts'
 import { filecoinCalibration, filecoin } from 'viem/chains'
 
-/**
- * @param {{
- *   ENVIRONMENT: 'mainnet' | 'calibration' | 'dev'
- *   RPC_URL: string
- *   FILBEAM_CONTROLLER_PRIVATE_KEY: `0x${string}`
- * }} env
- */
+/** @param {Env} env */
 export function getChainClient(env) {
   const chain = env.ENVIRONMENT === 'mainnet' ? filecoin : filecoinCalibration
   const transport = http(env.RPC_URL)


juliangruber (changes requested):

db/migrations/0017_add_settle_up_to_epoch.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE data_sets ADD COLUMN settle_up_to_epoch INTEGER;
```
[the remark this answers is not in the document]
juliangruber: We also need to do our own accounting and projections, so anything that's wonky here can lead to headaches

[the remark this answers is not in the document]
pyropy: > Are you happy to accept a PR by me making this logic tighter?
> 
> Payment logic has to be super tight, because if it isn't, and people notice, there will be a big drama

Feel free to add / suggest any improvements, I'd appreciate it.

payment-settler/lib/rail-settlement.js
```diff
@@ -0,0 +1,21 @@
...
+ */
+export async function getDataSetsForSettlement(db) {
+  const result = await db
+    .prepare(
+      `
+      SELECT id
+      FROM data_sets
+      WHERE (with_cdn = 1 OR lockup_unlocks_at >= datetime('now'))
+        AND terminate_service_tx_hash IS NULL
+        AND usage_reported_until >= datetime('now', '-30 days')
```
bajtos: Can you please confirm that this query will ignore data sets paid by sanctioned addresses?

If we don't exclude sanctioned payers when deciding which data sets to settle, then I don't see a reason to provide a list of data sets to settle - the smart contract can iterate over all data sets on our behalf.

See the discussion in https://github.com/filbeam/contracts/pull/1#discussion_r2428900449

payment-settler/bin/payment-settler.js
```diff
@@ -0,0 +1,134 @@
...
+        `Found ${dataSetIds.length} data sets for settlement:`,
+        dataSetIds,
+      )
+
+      const { request } = await publicClient.simulateContract({
+        account,
+        abi,
+        address: env.FILBEAM_CONTRACT_ADDRESS,
+        functionName: 'settleCDNPaymentRailBatch',
+        args: dataSetIds.map((id) => BigInt(id)),
```
bajtos: We will eventually need to split this TX into multiple smaller ones, when the list of data sets becomes too large.

I think our situation is simpler in the payment-settler worker, because we don't have to settle all data sets at once. We can settle up to X data sets per worker invocation, where X is chosen so that we don't exceed the max calldata size. In the next run, the worker will pick next X data sets that haven't been settled yet, and so on.

Anyhow, we don't need that yet.

Please create a follow-up issue and add a code comment here to remind us about the limitation.

payment-settler/test/worker.test.js
```diff
@@ -0,0 +1,375 @@
...
+import { withDataSet, randomId, getDaysAgo } from './test-helpers.js'
+import worker from '../bin/payment-settler.js'
+
+describe('payment settler scheduled handler', () => {
+  let mockPublicClient
+  let mockWalletClient
+  let mockAccount
+  let mockGetChainClient
+  let simulateContractCalls
+  let writeContractCalls
```
bajtos: Please add JSDoc comments to define types for these variables, so that we avoid `any` type.

indexer/lib/fwss-handlers.js
```diff
@@ -64,18 +65,40 @@ export async function handleFWSSDataSetCreated(
...
+ *   DB: D1Database
+ * }} env
+ * @param {object} payload
+ * @param {string} payload.data_set_id
+ * @param {string} payload.block_number
  * @throws {Error}
  */
 export async function handleFWSSServiceTerminated(env, payload) {
+  const DEFAULT_LOCKUP_PERIOD_DAYS = env.DEFAULT_LOCKUP_PERIOD_DAYS || 30 // 30 days default
+  const GENESIS_BLOCK_TIMESTAMP = env.GENESIS_BLOCK_TIMESTAMP || 1667326380 // Calibration genesis as fallback
```
juliangruber: This env value must always be configured, and the service mustn't startup if it isn't. A fallback is dangerous here

indexer/lib/fwss-handlers.js
```diff
@@ -64,18 +65,40 @@ export async function handleFWSSDataSetCreated(
...
+
+  // Convert block_number (epoch) to Unix timestamp (in seconds)
+  const epochTimestamp = epochToTimestamp(
+    payload.block_number,
+    GENESIS_BLOCK_TIMESTAMP,
+  )
```
juliangruber: Why are we needing timestamps here? Couldn't we only use black numbers everywhere exclusively, and not even care about how long one block is (or should be)?

juliangruber (changes requested):

indexer/lib/fwss-handlers.js
pyropy: I added asserts to indexer `fetch` method

https://github.com/filbeam/worker/pull/324/commits/f3084d58775c0b46c404e7a984f7efda7a209b59

payment-settler/bin/payment-settler.js
```diff
@@ -0,0 +1,134 @@
...
+ *   transactionHash: string
+ * }} TransactionRetryMessage
+ */
+
+/**
+ * @typedef {{
+ *   ENVIRONMENT: 'dev' | 'calibration' | 'mainnet'
+ *   RPC_URL: string
+ *   FILBEAM_CONTRACT_ADDRESS: string
+ *   FILBEAM_CONTROLLER_ADDRESS_PRIVATE_KEY: string
```
bajtos: Cross-posting my question from https://github.com/filbeam/roadmap/issues/87#issuecomment-3421650448:

Do you remember if we use a regular private-key-based address or a Safe-based multisig as the entity allowed to settle CDN payment rails? If we use Safe, then I think we need to slightly change how the settlement works.

/cc @juliangruber

payment-settler/lib/rail-settlement.js
pyropy: > Can you please confirm that this query will ignore data sets paid by sanctioned addresses?

There are multiple tests that confirm that data sets with `terminate_service_tx_hash` set will be excluded from the list returned by `getDataSetsForSettlement`.

> What's not clear to me: how do we differentiate between payment rails that were terminated by the client (e.g. when they terminate the entire data set) and the payment rails where we terminated the CDN service because the client was sanctioned?

Data sets terminated by the client will only have `with_cdn` flag set to `false` while those terminated by our `terminator` worker will have `terminate_service_tx_hash` column equal to the transaction in which we have terminated the service for the data set. 

> Also, what happens if the client terminates the data set just before we discover that they are sanctioned?

In your scenario we will still not try to settle payments for that data set as the `terminate_service_tx_hash` column has a set value. 

In some other cases where client terminates the service and we process the termination events and only after that we discover that user is sanctioned we will not submit the termination transaction and hence `terminate_service_tx_hash` column will not be populated for that data set.

> Have you considered joining data_sets with wallet_details and filtering out all data sets where the payer_address has the is_sanctioned flag set?

We might want to do that but we need to figure out if we want to settle payments for a payer that happened prior to them being sanctioned?

payment-settler/bin/payment-settler.js
pyropy: [Follow-up issue](https://github.com/filbeam/worker/issues/353)

indexer/bin/indexer.js
```diff
@@ -40,6 +41,9 @@ export default {
       return new Response('Unauthorized', { status: 401 })
     }
 
+    assert(env.DEFAULT_LOCKUP_PERIOD_DAYS)
+    assert(env.FILECOIN_GENESIS_BLOCK_TIMESTAMP_MS)
```
juliangruber: aren't those values part of the `env` type and therefore don't need to be asserted?

indexer/lib/fwss-handlers.js
```diff
@@ -64,18 +65,36 @@ export async function handleFWSSDataSetCreated(
 /**
  * Handle Filecoin Warm Storage Service service termination
  *
- * @param {{ DB: D1Database }} env
+ * @param {{
+ *   DB: D1Database
+ *   DEFAULT_LOCKUP_PERIOD_DAYS: number
+ *   FILECOIN_GENESIS_BLOCK_TIMESTAMP_MS: number
+ * }} env
```
juliangruber: with the new TS setup, isn't just `Env` enough here?

indexer/lib/fwss-handlers.js
```diff
@@ -64,18 +65,36 @@ export async function handleFWSSDataSetCreated(
...
  * @param {any} payload
  * @throws {Error}
  */
 export async function handleFWSSServiceTerminated(env, payload) {
+  // Convert block_number (epoch) to Unix timestamp (in milliseconds)
```
juliangruber:
```suggestion
```

indexer/lib/fwss-handlers.js
```diff
@@ -64,18 +65,36 @@ export async function handleFWSSDataSetCreated(
...
+  const epochTimestampMs = epochToTimestampMs(
+    payload.block_number,
+    Number(env.FILECOIN_GENESIS_BLOCK_TIMESTAMP_MS),
+  )
+
+  // Calculate lockup unlock timestamp based on the epoch timestamp (in milliseconds)
+  const lockupUnlocksAtMs =
+    epochTimestampMs +
+    Number(env.DEFAULT_LOCKUP_PERIOD_DAYS) * 24 * 60 * 60 * 1000
```
juliangruber: are we concerned that `DEFAULT_LOCKUP_PERIOD_DAYS` can get out of sync with the contract? If so, can we read it from it instead of from env?

payment-settler/lib/queue-handlers.js
juliangruber: It will be very good to deduplicate this soon. Could you please open a tech debt issue about it?

juliangruber (changes requested):

indexer/lib/fwss-handlers.js
bajtos: IIRC, we agreed with @pyropy to use the constant so that we can ship this sooner. `DEFAULT_LOCKUP_PERIOD_DAYS` is unlikely to change often.

How about creating a follow-up tech-debt issue for addressing this later? If we feel strongly about this, we can add the new issue to the GA launch backlog.

payment-settler/bin/payment-settler.js
```diff
@@ -0,0 +1,129 @@
...
+/**
+ * @typedef {{
+ *   ENVIRONMENT: 'dev' | 'calibration' | 'mainnet'
+ *   RPC_URL: string
+ *   FILBEAM_CONTRACT_ADDRESS: `0x${string}`
+ *   FILBEAM_CONTROLLER_ADDRESS_PRIVATE_KEY: `0x${string}`
+ *   DB: D1Database
+ *   TRANSACTION_MONITOR_WORKFLOW: Workflow
+ *   TRANSACTION_QUEUE: Queue<TransactionRetryMessage>
+ * }} RailSettlementEnv
```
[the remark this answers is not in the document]
bajtos: Yes, not necessary now.

Please use the `Env` type provided in the auto-generated `.d.ts` file.

payment-settler/lib/queue-handlers.js
bajtos: I think it's this one: https://github.com/filbeam/worker/issues/369

payment-settler/wrangler.toml
```diff
@@ -0,0 +1,112 @@
...
+
+[[env.dev.workflows]]
+binding = "TRANSACTION_MONITOR_WORKFLOW"
+name = "dev-transaction-monitor-workflow"
+class_name = "TransactionMonitorWorkflow"
+
+[env.calibration.vars]
+ENVIRONMENT = "calibration"
+RPC_URL = "https://api.calibration.node.glif.io/"
+FILBEAM_CONTRACT_ADDRESS = "0x"
```
bajtos: Missing contract address.

Should we wait with landing this PR until there is a FilBeamOperator version deployed to calibration?

Alternatively, how about changing the cron schedule so that this worker is never executed?

payment-settler/lib/epoch.js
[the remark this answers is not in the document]
pyropy: I'd like to ship it first and refactor later if possible 🙏🏻

payment-settler/wrangler.toml
```diff
@@ -0,0 +1,112 @@
...
+name = "calibration-transaction-monitor-workflow"
+class_name = "TransactionMonitorWorkflow"
+
+[env.calibration.triggers]
+crons = ["*/1 * * * *"]  # Run every minute
+
+[env.mainnet.vars]
+ENVIRONMENT = "mainnet"
+RPC_URL = "https://api.node.glif.io/"
+FILBEAM_CONTRACT_ADDRESS = "0x" # TODO
```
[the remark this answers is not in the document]
juliangruber: I've removed the schedule trigger and added a todo to the canvas

payment-settler/package.json
```diff
@@ -0,0 +1,24 @@
...
+    "start": "wrangler d1 migrations apply dev-db --local --env dev --cwd ../db  && wrangler dev --env dev",
+    "test": "vitest run"
+  },
+  "dependencies": {
+    "@filbeam/workflows": "^1.0.0",
+    "viem": "^2.37.9"
+  },
+  "devDependencies": {
+    "@cloudflare/vitest-pool-workers": "^0.9.7",
+    "@types/validator": "^13.15.2"
```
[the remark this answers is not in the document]
juliangruber: 0e2d0a9

PR #333, opened 2025-10-06
feat: update egress quota based on CDN payment rail changes

pyropy:
## Summary
- Implement egress quota calculation when CDNPaymentRailsToppedUp event is emitted
- Store quotas in database for each data set
- Configure egress rate in worker configuration (e.g., $5 per TiB)

## Changes
- Added database migration for quota columns (cdn_egress_quota, cache_miss_egress_quota)
- Added handler for CDNPaymentRailsToppedUp events
- Updated CDNPaymentRailsToppedUp event in subgraph to include new fields (cdnAmountAdded, cacheMissAmountAdded)
- Refactored quota calculation to use amounts added from event instead of total amounts
- Updated indexer to calculate new egress quota based on incremental amounts

Related to:
- https://github.com/filbeam/roadmap/issues/11
- https://github.com/filbeam/roadmap/issues/54
- https://github.com/FilOzone/filecoin-services/pull/284 (blocker)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5640,9 +5640,9 @@
       }
     },
     "node_modules/magic-string": {
-      "version": "0.30.19",
-      "resolved": "https://registry.npmjs.org/magic-string/-/magic-string-0.30.19.tgz",
-      "integrity": "sha512-2N21sPY9Ws53PZvsEpVtNuSW+ScYbQdp4b9qUaL+9QkHUrGFKo56Lg9Emg5s9V/qrtNBmiR01sYhUOwu3H+VOw==",
+      "version": "0.30.21",
+      "resolved": "https://registry.npmjs.org/magic-string/-/magic-string-0.30.21.tgz",
+      "integrity": "sha512-vd2F4YUyEXKGcLHoq+TEyCjxueSeHnFxyyjNp80yg0XV4vUhnDer/lvvlqM/arB5bXQN5K2/3oinyCRyx8T2CQ==",
       "dev": true,
       "license": "MIT",
       "dependencies": {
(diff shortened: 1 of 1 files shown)


indexer/test/fwss-cdn-payment-rails.test.js
bajtos: We should use numeric IDs to verify that our SQL queries are correctly converting them to strings.
  
```
dataSetId: '100',
```         

We had issues in the past where CF was storing numbers as floats (e.g. 100.0) and equality checks were not working correctly in our queries.

On second thought, why do we need a specific dataSetId value here? Can we let the test data builder generate one?

indexer/lib/fwss-handlers.js
```diff
@@ -126,7 +121,7 @@ export async function handleFWSSCDNPaymentRailsToppedUp(env, payload) {
     .bind(
       Number(cdnEgressQuotaAdded),
       Number(cacheMissEgressQuotaAdded),
-      dataSetId,
+      payload.data_set_id,
```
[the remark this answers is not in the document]
bajtos: Oh, I see this was discussed in https://github.com/filbeam/worker/pull/333#discussion_r2428455686

I am concerned whether we have enough safety mechanisms in place to prevent us from accidentally calling `handleFWSSCDNPaymentRailsToppedUp` with a numeric ID or a numeric amount.

Having said that, I see that the method has JSDoc clearly specifying that the payload properties should have string type, so I think we are good 👍🏻 

Please ignore my comment above; no changes are needed here.

indexer/test/fwss-cdn-payment-rails.test.js
bajtos: Based on the discussion in other comments, where I learned that payload's data set ID is always string, I guess that it's ok to keep using a non-numeric string ID in the tests.

bajtos (approved):

bajtos:
@juliangruber I think this is waiting for your review, can you PTAL?

pyropy:
> The diff has unrelated parts, could you please update the PR?

I've pulled changes from #316 and it resolved the issue

PR #354, opened 2025-10-14
piece-retriever: fix Request#tee() buffer exceeded

juliangruber:
We were caching the whole value before sending the tee() clone to the client, leading to buffer exceeded issues.


juliangruber:
I have been unable to create a test for this. I tried constructing a response body of 70mb size, which we know would fail in production currently, and consuming it, but it didn't create an error. I suspect we would need to consume slowly, for the buffer to fill up, but then the test would take long to pass. @bajtos do you think it's still worth it to try to follow with a PR for this strategy?

PR #357, opened 2025-10-15
refactor: Extract workflows into a  package

pyropy:
## Summary

Extracts the `TransactionMonitorWorkflow` into a standalone `@filbeam/workflows` package to enable reuse across multiple workers (terminator and usage-reporter).

## Changes

- Moved `TransactionMonitorWorkflow` from `terminator/lib/` to `workflows/lib/`
- Updated terminator to import workflow from the new package (`@filbeam/workflows`)

### DIFF
--- a/piece-retriever/bin/piece-retriever.js
+++ b/piece-retriever/bin/piece-retriever.js
@@ -101,7 +101,11 @@ export default {
         `Unsupported Service Provider: ${serviceProviderId}`,
       )
 
-      const { response: originResponse, cacheMiss } = await retrieveFile(
+      const {
+        response: originResponse,
+        cacheMiss,
+        url,
+      } = await retrieveFile(
         ctx,
         serviceUrl,
         pieceCid,
@@ -109,6 +113,30 @@ export default {
         env.ORIGIN_CACHE_TTL,
         { signal: request.signal },
       )
+
+      if (originResponse.status >= 500) {
+        ctx.waitUntil(
+          logRetrievalResult(env, {
+            cacheMiss,
+            responseStatus: originResponse.status,
+            egressBytes: 0,
+            requestCountryCode,
+            timestamp: requestTimestamp,
+            dataSetId,
+          }),
+        )
+        const response = new Response(
+          `Service provider ${serviceProviderId} is unavailable at ${url}`,
+          {
+            status: 502,
+            headers: new Headers({
+              'X-Data-Set-ID': dataSetId,
+            }),
+          },
+        )
+        setContentSecurityPolicy(response)
+        return response
+      }
 
       if (!originResponse.body) {
         // The upstream response does not have any readable body
--- a/piece-retriever/lib/retrieval.js
+++ b/piece-retriever/lib/retrieval.js
@@ -12,8 +12,8 @@
  * @returns {Promise<{
  *   response: Response
  *   cacheMiss: boolean
+ *   url: string
  * }>}
- *
  *   - The response from the fetch request, the cache miss and the content length.
  */
 export async function retrieveFile(
@@ -52,7 +52,7 @@ export async function retrieveFile(
     }
   }
 
-  return { response, cacheMiss }
+  return { response, cacheMiss, url }
 }
 
 /**
--- a/piece-retriever/test/retrieval.test.js
+++ b/piece-retriever/test/retrieval.test.js
@@ -23,12 +23,16 @@ describe('retrieveFile', () => {
   it('constructs the correct URL', async () => {
     cachesMock.match.mockResolvedValueOnce(null)
     const ctx = createExecutionContext()
-    await retrieveFile(ctx, baseUrl, pieceCid, new Request(baseUrl))
-    await waitOnExecutionContext(ctx)
-    expect(fetchMock).toHaveBeenCalledWith(
-      `${baseUrl}/piece/${pieceCid}`,
-      expect.any(Object),
+    const { url } = await retrieveFile(
+      ctx,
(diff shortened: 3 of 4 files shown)


workflows/package.json
```diff
@@ -0,0 +1,21 @@
...
+  "name": "@filbeam/workflows",
+  "version": "1.0.0",
+  "private": true,
+  "description": "Shared workflows for FilBeam workers",
+  "author": "Space Meridian <filbeam@meridian.space>",
+  "type": "module",
+  "exports": {
+    ".": "./lib/transaction-monitor-workflow.js"
+  },
+  "main": "lib/transaction-monitor-workflow.js",
```
bajtos: Why do we need the `exports` block when we already have the `main` entry point specified? Would the following work?


```suggestion
  "main": "lib/transaction-monitor-workflow.js",
```

workflows/wrangler.toml
bajtos: Why is this wrangler configuration file needed in a shared library that we are not deploying to Cloudflare? What happens if we remove this file?

workflows/test/transaction-monitor-workflow.test.js
```diff
@@ -0,0 +1,341 @@
...
+      RPC_URL: 'https://api.calibration.node.glif.io/rpc/v0',
+      FILCDN_CONTROLLER_ADDRESS_PRIVATE_KEY: '0x1234',
+    }
+
+    mockStep = {
+      do: vi.fn(),
+    }
+
+    workflow = Object.create(TransactionMonitorWorkflow.prototype)
+    workflow.env = mockEnv
```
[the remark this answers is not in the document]
pyropy: I attempted this approach when I first introduced workflows, but couldn't get it working correctly. Creating a new `TransactionMonitorWorkflow` instance this way resulted in an error, even after trying to pass the context returned by `createExecutionContext`.

```js
TypeError: Failed to construct 'WorkflowEntrypoint': constructor parameter 1 is not of type 'ExecutionContext'.
 ❯ new TransactionMonitorWorkflow lib/transaction-monitor-workflow.js:20:8
     18|  * upToTimestamp }, retryData: { upToTimestamp } } } })
     19|  */
     20| export class TransactionMonitorWorkflow extends WorkflowEntrypoint {
       |        ^
     21|   /**
     22|    * @param {WorkflowEvent} event
```

workflows/wrangler.toml
pyropy: Unfortunately, we still need to use the workers package from Wrangler during testing, since our `TransactionMonitorWorkflow` extends `WorkflowEntrypoint`. I haven't found package on the npm.

workflows/test/transaction-monitor-workflow.test.js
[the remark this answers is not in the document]
bajtos: OK, I think we don't need perfect tests right now and should focus on shipping this ASAP.

I propose the following:

- Keep your current solution that works
- Create a follow-up tech-debt issue to rework these tests to use the canonical CF testing approach (see the examples I pointed to)
- Add a code comment mentioning this problem and pointing to the tech-debt issue

WDYT?

workflows/test/transaction-monitor-workflow.test.js
```diff
@@ -0,0 +1,341 @@
...
+
+vi.mock('../lib/chain.js', () => ({
+  getChainClient: vi.fn(),
+}))
+
+describe('TransactionMonitorWorkflow', () => {
+  let workflow
+  let mockEnv
+  let mockStep
+  let mockQueue
```
bajtos: These variables have type `any`, which disables type checks. Let's tell TypeScript about their types.


```suggestion
  /** @type {TransactionMonitorWorkflow} */
  let workflow
  /**
   * @type {{
   *   TRANSACTION_QUEUE: {
   *     send: Mock<Procedure>
   *   }
   *   ENVIRONMENT: string
   *   RPC_URL: string
   *   FILCDN_CONTROLLER_ADDRESS_PRIVATE_KEY: string
   * }}
   */
  let mockEnv
  /**
   * @type {{
   *   do: Mock<Procedure>
   * }}
   */
  let mockStep
  /**
   * @type {{
   *   send: Mock<Procedure>
   * }}
   */  
  let mockQueue
```

workflows/wrangler.toml
```diff
@@ -0,0 +1,11 @@
+name = "workflows"
```
bajtos:
```suggestion
name = "filbeam-workflows"
```

workflows/test/transaction-monitor-workflow.test.js
pyropy: Added in https://github.com/filbeam/worker/pull/357/commits/e46a0a728d06ba5ac65aceab099c36396a94c50d

workflows/test/transaction-monitor-workflow.test.js
```diff
@@ -0,0 +1,362 @@
...
+import { TransactionMonitorWorkflow } from '../lib/transaction-monitor-workflow.js'
+
+vi.mock('../lib/chain.js', () => ({
+  getChainClient: vi.fn(),
+}))
+
+describe('TransactionMonitorWorkflow', () => {
+  /** @type {WorkflowStep} */
+  const mockStep = { do: vi.fn() }
+  /** @type {{ send: (message: any) => Promise<void> }} */
```
bajtos: Please avoid `any`. You can use `unknown` to allow any message type.


```suggestion
  /** @type {{ send: (message: unknown) => Promise<void> }} */
```

PR #374, opened 2025-10-22
chore: setup TypeScript project references + fix tsc errors

bajtos:
- **refactor: extract tsconfig.base.json**
- **chore: setup TypeScript project references**
- **chore: fix TypeScript errors**
- **chore: bump tsconfig libs to `esnext`**

Also: upgrade compatibility date to "2024-09-08". (We cannot use a newer date due to https://github.com/cloudflare/workers-sdk/issues/11028.)

Fixes https://github.com/filbeam/worker/issues/368


juliangruber (changes requested):
Massive simplification! And as it looks, will get us better TS validation too 👍

terminator/lib/queue-handlers.js
```diff
@@ -150,24 +139,33 @@ export async function handleTransactionRetryQueueMessage(
       `Calculating gas fees from the recent Send message ${recentSendMessage.cid}`,
     )
 
+    const originalMaxPriorityFeePerGas = originalTx.maxPriorityFeePerGas
+    assert(
+      originalMaxPriorityFeePerGas !== undefined,
+      'originalTx.maxPriorityFeePerGas is null',
+    )
```
bajtos: According to TypeScript typings for viem, `originalTx.maxPriorityFeePerGas` may be undefined.

Do we know if that's actually possible and under which conditions it would happen? Do we need to handle this case or is it OK to keep my assert? What would happen when the assertion fails - will we drop the TX, keep retrying it forever, something else?

cc @pyropy

terminator/lib/queue-handlers.js
```diff
@@ -150,24 +139,33 @@ export async function handleTransactionRetryQueueMessage(
...
 
     const newGasLimit = BigInt(
       Math.min(
         Math.ceil(
-          Math.max(Number(originalTx.gasLimit), recentSendMessage.gasLimit) *
-            1.1,
+          // THIS REQUIRES CAREFUL REVIEW
+          // Viem does not provide `originalTx.gasLimit`, only `gas`.
+          // Is this new calculation still correct?
+          Math.max(Number(originalTx.gas), recentSendMessage.gasLimit) * 1.1,
```
bajtos: Viem does not provide `originalTx.gasLimit`, only `gas`. Is this new calculation still correct?

@pyropy Do you know what the correct solution is here?

terminator/lib/queue-handlers.js
```diff
@@ -150,24 +139,33 @@ export async function handleTransactionRetryQueueMessage(
...
         ),
         1e10, // block gas limit
       ),
     )
 
+    const recentGasFeeCap = BigInt(recentSendMessage.gasFeeCap)
```
bajtos: `recentSendMessage.gasFeeCap` is a string, not a bigint.

workflows/lib/transaction-monitor-workflow.js
```diff
@@ -32,6 +39,7 @@ export class TransactionMonitorWorkflow extends WorkflowEntrypoint {
         {
           timeout: `5 minutes`,
           retries: {
+            delay: '10 seconds',
             limit: 3,
           },
```
bajtos: `delay` is required when configuring retries. Is `10 seconds` a reasonable value or should we change it to something else?

@pyropy

README.md
```diff
@@ -12,6 +12,13 @@ Create `indexer/.dev.vars` file with the following content:
...
 SECRET_HEADER_KEY=X-SECRET-KEY
 SECRET_HEADER_VALUE=SecretToken
+CHAINALYSIS_API_KEY=YourChainalysisApiKey
+```
+
+Create `terminator/.dev.vars` file with the following content:
+
+```
+FILCDN_CONTROLLER_ADDRESS_PRIVATE_KEY=0xSomePrivateKey
 ```
```
bajtos: It is key to have these files present when running `npm run build:types`, as that's the way how `wrangler` discovers what secrets are available in the environment.

We often forget to create these files. When that happens the next time (after my PR was landed), we will get TypeScript compiler errors flagging that secrets are not available on the `Env` object. That could be confusing, as it's not clear how to fix it.

How can we improve our DX here? I can imagine creating a script to populate `.dev.vars` files with some sensible defaults, WDYT? But then we need to remember to update that script whenever we add a new secret 🤔

package.json
```diff
@@ -16,19 +16,21 @@
     "build:types": "npm run build:types --workspaces --if-present",
     "deploy:calibration": "npm run deploy:calibration --workspaces --if-present",
     "deploy:mainnet": "npm run deploy:mainnet --workspaces --if-present",
-    "lint": "eslint && prettier --check . && tsc -p .",
+    "lint": "eslint && prettier --check . && tsc -b",
     "lint:fix": "eslint --fix && prettier --write .",
-    "test": "npm run lint && wrangler d1 migrations apply test-db --local --cwd db && vitest run"
+    "test": "npm run lint && wrangler d1 migrations apply test-db --local --cwd db && vitest run",
+    "update-ts-project": "node bin/update-ts-project.js"
```
bajtos: We need to run this command every time we add a new workspace package or add a new dependency between workspace packages. I am sure we will often forget to do that.

How can we help us remember?


I have two options in my mind:

1. Run `update-ts-project` every time we run `npm lint`. That will update tsconfig files, which is IMO enough to remind us we need to update references. Downsides: most of the time, the script does not change anything, so it's adding unnecessary delay.
2. Modify the script updating project references - introduce a `--check` mode. The script should not change anything, just report whether project references are up to date. (Exit code 0: all is good. Exit code 1: updates are needed.) Then we can add a CI step to run that check. Downside: a bit more work to do here.

WDYT? Do you like any of those options? Can you suggest a different approach?

I propose to leave this improvement out of the scope of this pull request, so that we can land the core changes faster & avoid merge conflicts.

subgraph/tsconfig.json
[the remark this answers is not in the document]
bajtos: This was accidental; I restored the file.

terminator/bin/terminator.js
```diff
@@ -16,31 +15,14 @@ import {
 /**
  * @typedef {{
  *   type: 'transaction-retry'
- *   transactionHash: string
+ *   transactionHash: `0x${string}`
```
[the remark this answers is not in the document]
bajtos: Here is the tsc error message when I revert the change on this line:

```
terminator/bin/terminator.js:69:54 - error TS2345: Argument of type 'import("/Users/bajtos/src/filbeam/worker/terminator/bin/terminator").TransactionRetryMessage' is not assignable to parameter of type 'import("/Users/bajtos/src/filbeam/worker/terminator/lib/queue-handlers").TransactionRetryMessage'.
  Types of property 'transactionHash' are incompatible.
    Type 'string' is not assignable to type '`0x${string}`'.

69             await handleTransactionRetryQueueMessage(message.body, env)
                                                        ~~~~~~~~~~~~


Found 1 error.
```

The important part is this:

```
  Types of property 'transactionHash' are incompatible.
    Type 'string' is not assignable to type '`0x${string}`
```

Viem has stricter types; it requires TX hashes and addresses to start with a `0x` prefix, as is common in the Ethereum world.

Cloudflare generates `string` type for environment variables and secrets.

terminator/lib/queue-handlers.js
```diff
@@ -150,24 +150,33 @@ export async function handleTransactionRetryQueueMessage(
...
 
     const newGasLimit = BigInt(
       Math.min(
         Math.ceil(
-          Math.max(Number(originalTx.gasLimit), recentSendMessage.gasLimit) *
-            1.1,
+          // THIS REQUIRES CAREFUL REVIEW
+          // Viem does not provide `originalTx.gasLimit`, only `gas`.
+          // Is this new calculation still correct?
+          Math.max(Number(originalTx.gas), recentSendMessage.gasLimit) * 1.1,
```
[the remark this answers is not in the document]
bajtos: Let's discuss in https://github.com/filbeam/worker/pull/374#discussion_r2451312130, please.

tsconfig.base.json
```diff
@@ -0,0 +1,20 @@
+{
+  "$schema": "https://json.schemastore.org/tsconfig",
+  "compilerOptions": {
+    "allowJs": true,
+    "checkJs": true,
+    "skipLibCheck": true,
+    "lib": ["es2022"],
+    "target": "es2022",
```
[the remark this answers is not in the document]
bajtos: Done in [727c3ea](https://github.com/filbeam/worker/pull/374/commits/727c3eaaf9b70c7dadd6ed891cd6cc67c81b234e)

workflows/lib/transaction-monitor-workflow.js
pyropy: 10 seconds sounds good to me! 👍🏻

PR #394, opened 2025-10-27
fix: await `getBadBitsEntry()`

juliangruber:
Bad bits wasn't working at all, since we just checked for `Promise (pending)`

### DIFF
--- a/piece-retriever/bin/piece-retriever.js
+++ b/piece-retriever/bin/piece-retriever.js
@@ -74,7 +74,7 @@ export default {
       const [{ serviceProviderId, serviceUrl, dataSetId }, isBadBit] =
         await Promise.all([
           getStorageProviderAndValidatePayer(env, payerWalletAddress, pieceCid),
-          env.BAD_BITS_KV.get(`bad-bits:${getBadBitsEntry(pieceCid)}`, {
+          env.BAD_BITS_KV.get(`bad-bits:${await getBadBitsEntry(pieceCid)}`, {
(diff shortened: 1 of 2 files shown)


piece-retriever/bin/piece-retriever.js
```diff
@@ -74,7 +74,7 @@ export default {
       const [{ serviceProviderId, serviceUrl, dataSetId }, isBadBit] =
         await Promise.all([
           getStorageProviderAndValidatePayer(env, payerWalletAddress, pieceCid),
-          env.BAD_BITS_KV.get(`bad-bits:${getBadBitsEntry(pieceCid)}`, {
+          env.BAD_BITS_KV.get(`bad-bits:${await getBadBitsEntry(pieceCid)}`, {
```
bajtos: Nitpick: this will await `getBadBitsEntry` first, and only then start `getStorageProviderAndValidatePayer` and `env.BAD_BITS_KV.get` in parallel.

IMO, a better solution would be to refactor `getBadBitsEntry` and `env.BAD_BITS_KV.get` into a standalone function, so that both legs can run concurrently from the beginning.

It makes me wonder if such a change would have a measurable impact on the retrieval response times 🤔 

See also `findInBadBits(env, cid)` we used to use before:
https://github.com/filbeam/worker/blob/e5681315b6d81d542d51966710cef3cca825a2d4/piece-retriever/lib/bad-bits-util.js#L19-L27

PR #412, opened 2025-10-30
piece-retrieval: add `X-Cache` header

juliangruber:
Useful for debugging


bajtos:
Great idea!

It makes me wonder how `X-Cache` is different from the response headers added by Cloudflare, and how to make that distinction easier to understand to FilBeam users.

Perhaps `FilBeam-Cache-Status` would be a better header name?

https://developers.cloudflare.com/cache/concepts/cache-responses/

PR #431, opened 2025-11-05
piece-retriever: add pick random SP

juliangruber:
Since we will always have 2 SPs per data set, distribute traffic ~evenly.

Future steps:
1. Retry with different SP if picked one failed (this is kinda obvious)
2. Sort by remaining egress descending

Not so sure about the 2nd one, it can make it so SPs with higher egress budget can starve small SPs.


pyropy (approved):
Great work, Julian! 🎉

PR #451, opened 2025-11-07
fix: workflow id too long

pyropy:
Fixes errors caused workflow ID being too long. Workflow ID is capped at a maximum of 100 characters. These errors are emitted during workflow creation by the [workflow validation function](https://github.com/cloudflare/workers-sdk/blob/4937e85a0df6ffd86b1730623012dcb9ee31127c/packages/workflows-shared/src/lib/validators.ts#L27-L38).


pyropy:
@juliangruber @bajtos Please make a retroactive pull-request review 🙏🏻

PR #469, opened 2025-11-13
fix: charge bots for traffic

bajtos:
This partially reverts e664eb6.


juliangruber (approved):

bajtos:
> Could you also revert changes made in #381 ?

Which changes do you mean?

I find it useful to keep track of which requests are coming from which bot.

Are you suggesting to stop tracking bots and remove the auth layer added by #381 completely?

Issue #305, opened 2025-09-26
Verify integrity of content returned by SP


bajtos:
Our retriever worker needs to verify the integrity of the data returned by the SP. The content bytes must match the requested PieceCID.

When the hash does not match, we need to flag this in our stats and don't reward SP for the cache-miss egress.

Our dashboard(s) should indicate which SPs are returning content that does not match PieceCID.

juliangruber:
- [ ] Don't count invalid responses as cache miss egress in worker: https://github.com/filbeam/worker/pull/498
- [ ] Don't count invalid responses as cache miss egress in dashboard: https://github.com/filbeam/dashboard/pull/29
- [ ] Expose invalid cache miss responses in dashboard

PR #508, opened 2025-12-10
perf: adopt WASM-based PieceCID hasher

bajtos:
This is a follow-up for #506 that's adding even more performant hasher implementation. See also #305.

- `node:crypto` version: 3153 KB/s
- WASM version: 4976 KB/s


juliangruber (approved):
Awesome!!!

Issue #507, opened 2025-12-10
Implement X402 Paywall for FilBeam (PoC)


pyropy:
### Overview

Implement X402 protocol support to enable content creators to monetize their content stored on Filecoin PDP. This PoC allows creators to set a price for piece retrieval, with payments processed via the X402 Facilitator on Base Network.

### Background

X402 is a retrieval protocol that enables content owners to charge clients for content access. This aligns with FilBeam's mission of solving retrieval incentives in decentralized storage networks.

**Why Base Network?** Base provides fast finality times, which is critical for our Cloudflare Workers environment and ensures a good user experience. This also enables listing in X402 community tooling.

### Acceptance Criteria

- Users are able to upload content with specified content price & owner (payout) address
- Requesters are able to pay to access the content using USDC on Base network
- Requesters are able to access the content after successful payment
- Content owner receives full payment amount

PR #518, opened 2025-12-15
feat: index x402 metadata

pyropy:
## Description

This PR enables indexing of X402 metadata. 

Indexed metadata is parsed during the piece insertion. Additionally, metadata is stored in a new X402 metadata KV store. This new metadata store should enable fast lookups for the new X402 retrieval worker. 

## Change log

- Add X402 price field to piece table
- Add X402 piece metadata KV
- Index X402 piece price to database and metadata KV
- On piece deletion delete metadata if no other pieces for the payer with the same CID are remaining

Related to #507

### DIFF
--- a/indexer/package.json
+++ b/indexer/package.json
@@ -19,7 +19,7 @@
     "validator": "^13.15.23"
   },
   "devDependencies": {
-    "@cloudflare/vitest-pool-workers": "^0.10.15",
+    "@cloudflare/vitest-pool-workers": "^0.11.0",
     "@types/validator": "^13.15.10"
   }
 }
--- a/package-lock.json
+++ b/package-lock.json
@@ -20,9 +20,9 @@
       ],
       "devDependencies": {
         "@checkernetwork/prettier-config": "^1.0.1",
-        "@cloudflare/vitest-pool-workers": "^0.10.15",
+        "@cloudflare/vitest-pool-workers": "^0.11.0",
         "@npmcli/map-workspaces": "^5.0.3",
-        "@types/node": "^25.0.2",
+        "@types/node": "^25.0.3",
         "neostandard": "^0.12.1",
         "prettier": "^3.7.4",
         "typescript": "^5.9.3",
@@ -47,7 +47,7 @@
         "validator": "^13.15.23"
       },
       "devDependencies": {
-        "@cloudflare/vitest-pool-workers": "^0.10.15",
+        "@cloudflare/vitest-pool-workers": "^0.11.0",
         "@types/validator": "^13.15.10"
       }
     },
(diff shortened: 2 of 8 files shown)


indexer/lib/pdp-verifier-handlers.js
```diff
@@ -1,49 +1,128 @@
...
+  const remainingPiecesCids = new Set(remainingPieces.results.map((r) => r.cid))
+  const orphanedCids = deletedCids.filter(
+    (cid) => !remainingPiecesCids.has(cid),
+  )
+
+  await Promise.all(
+    orphanedCids.map((cid) =>
+      env.X402_METADATA_KV.delete(`${payerAddress}:${cid}`),
+    ),
+  )
```
bajtos: Consider the following scenario:

- piece `bafk123` is added with `x402Price:100` at block number `123`
- later, piece `bafk123`is added with `x402Price:200` at block number `256`
  - this updates the price in the KV cache from `100` to `200`
- even later, the piece with price `200` is removed, but the piece with price `100` remains

In the current implementation, we will keep charging `200` for access.

I think we should charge `100` for egress instead. Assuming the rule is "the most recent piece defines the price".

Thoughts?

If we agree to implement this change, then please remember to update the test(s) accordingly.

Also: Since this is a PoC, I think it's not necessary to handle all edge cases now. We can keep the current implementation and create a follow-up issue to fix the way how we handle piece deletion.

bajtos: When we get to implementing data set deletion (see https://github.com/filbeam/roadmap/issues/19), we will need to remember to update the x402 price in our KV cache as part of that process.

indexer/lib/pdp-verifier-handlers.js
```diff
@@ -1,49 +1,128 @@
...
+    .all()
+
+  const remainingPiecesCids = new Set(remainingPieces.results.map((r) => r.cid))
+  const orphanedCids = deletedCids.filter(
+    (cid) => !remainingPiecesCids.has(cid),
+  )
+
+  await Promise.all(
+    orphanedCids.map((cid) =>
+      env.X402_METADATA_KV.delete(`${payerAddress}:${cid}`),
```
bajtos: Note: We can make at most 1000 KV operations per worker invocation. This will be a problem if the list of orphaned CIDs ever grows to 1000 or more items.

https://developers.cloudflare.com/kv/platform/limits/

I think we don't need to worry about this edge until we encounter it.

Thoughts?

indexer/lib/pdp-verifier-handlers.js
```diff
@@ -1,49 +1,128 @@
...
+  const metadataKey = `${payerAddress}:${pieceCid}`
+
+  /** @type {{ price: string; blockNumber: number } | null} */
+  const existingMetadata = await env.X402_METADATA_KV.get(metadataKey, 'json')
+
+  if (existingMetadata && blockNumber < Number(existingMetadata.blockNumber)) {
+    return
+  }
+
+  await env.X402_METADATA_KV.put(
```
juliangruber: This write after read is not atomic. If we get multiple events in close proximity (eg during initial ingestion), likely, this will lead to inconsistent data. Do you have suggestions how to fix this?

pyropy: > This write after read is not atomic. If we get multiple events in close proximity (eg during initial ingestion), likely, this will lead to inconsistent data. Do you have suggestions how to fix this?

We're using block number to try and store the latest result in case of multiple out of order events. However, we're only applying this to the KV store and not the D1 database itself. There might be some inconsistency in case that same piece with different metadata is added to the same data set. 

> What if we store the x402 price in sql as well, in the piece table?

We do store the price in the SQL too.

My suggestion would be to also add a block column to pieces so we only update the piece row if the block number is bigger than the one stored in the database (similar for what we are doing in case of KV store).

indexer/lib/pdp-verifier-handlers.js
pyropy: Since we're not storing block numbers in piece row that would mean that we'd have to keep list of objects as the piece metadata (i.e. [{ id, blockNumber, price }]) and determine price in runtime by looking for the biggest block? 

Other implementation we could try is to have a block number in the piece row and find the price for biggest block number inside the D1 database.  WDYT?

indexer/lib/pdp-verifier-handlers.js
bajtos: Based on what we discussed in our call (choose the max price tag), we don't need to store the block number in the DB.

PR #530, opened 2025-12-23
fix: payment-settler calldata too big

### LINKED ISSUE #353: Split settlement call into multiple transactions
We will eventually need to handle the case when `[dataSetIdsBigInt]` is larger than max calldata size and trigger settlements in multiple TXs. Can you open a follow-up issue for that?

_Originally posted by @bajtos in https://github.com/filbeam/worker/pull/324#discussion_r2416150694_

pyropy:
## Description

Due to increasing number of data sets size of the payment settlement transactions exceeded the maximum transaction size. In order to fix this issue payment settlement is now performed in batches, where batch consists of up to 150 data sets. 

## Changelog

- Split data set settlement into multiple batches too avoid hitting calldata limit
- Added nonce manager to enable sending multiple transactions in parallel
- Spawn transaction monitoring workflows using the batch method

Closes #353


pyropy:
Hindsight please 🙏🏻 @juliangruber @bajtos

PR #537, opened 2026-01-05
perf: remove `response.body.tee()`

bajtos:
Let the built-in `fetch()` caching layer write the cache-miss response to the default cache.

Keep the explicit cache check for better cache-hit performance.

Note: I modified my original exploration in https://github.com/bajtos/cf-latencies to test this new caching setup. As far as I can tell, it works seamlessly.

Related pull requests:
- #315
- #354

My primary motivation for this pull request is to get rid of the following alerts:

```
ReadableStream.tee() buffer limit exceeded. This error usually occurs when a Request 
or Response with a large body is cloned, then only one of the clones is read, forcing 
the Workers runtime to buffer the entire body in memory. To fix this issue, remove 
unnecessary calls to Request/Response.clone() and ReadableStream.tee(), and always 
read clones/tees in parallel.
```

### DIFF
--- a/piece-retriever/lib/retrieval.js
+++ b/piece-retriever/lib/retrieval.js
@@ -40,7 +40,6 @@ export async function retrieveFile(
   if (response) {
     cacheMiss = false
   } else {
-    console.log(`Cache miss for URL: ${url}`)
     response = await fetch(url, {
       cf: {
         cacheEverything: true,


bajtos:
@juliangruber @pyropy hindsight please 🙏🏻

PR #538, opened 2026-01-06
fix: don't cache error cache-miss responses

bajtos:
This is a follow-up for https://github.com/filbeam/worker/pull/537, which performed a partial revert of https://github.com/filbeam/worker/pull/315 and did not restore the full `cf` fetch config.


juliangruber (commented):
+1

PR #554, opened 2026-01-15
Chore: Add agent files

pyropy:
Adds initial version of AGENTS.md file


bajtos:
> * Create a section with a check-list for adding a new worker, include a step to update these instructions
> * Tell Claude to run `npm run update-ts-projects` whenever a new package is added or a dependency on workspace-local package is added

I am adding these instructions in https://github.com/filbeam/worker/pull/560

PR #604, opened 2026-01-29
use filbeam contract in terminator n correct env names

### LINKED ISSUE #362: Proxy Filecoin warm storage calls through FilecoinBeamOperator and rename controller private key
Refactor the cloudflare workers to route all `FilecoinWarmStorageService` contract calls through the `FilecoinBeamOperator` contract as a proxy layer, and rename the environment variable from `FILCDN_CONTROLLER_ADDRESS_PRIVATE_KEY` to `FILBEAM_CONTROLLER_PRIVATE_KEY` for consistency.

DarkLord017:
fixes #362 

i just found contract calls in terminate directory are subgraphs and events also supposed to be proxied ?
Also i did not push worker configuration , was that supposed to be pushed ?

pls review @pyropy


workflows/lib/transaction-monitor-workflow.js
```diff
@@ -1,4 +1,9 @@
-/** @import {WorkflowEvent, WorkflowStep} from 'cloudflare:workers' */
+/**
```
pyropy: Let's revert this format.

pyropy (changes requested):
Great work! Let's just rename the `FILBEAM_CONTROLLER_PRIVATE_KEY` to `FILBEAM_OPERATOR_CONTROLLER_PRIVATE_KEY` since the contract name is `FilBeamOperator`.

pyropy (changes requested):
Please run `npm run lint:fix` as formatting is not consistent with the rest of the codebase.

rjan90:
Thanks for the PR. Unfortunately, we’ll need to close it for now. We can revisit, clean it up, and reopen it when there’s sufficient bandwidth and FilBeam is prioritized again.

PR #676, opened 2026-06-22
DRY: share the worker fetch wrapper

juliangruber:
Stacked on `serve-ipfs-retrievals`. Now that the `fetch` wrappers in `ipfs-retriever`, `piece-retriever` and `x402-piece-gateway` are nearly identical, this extracts the shared part into a `handleFetchRequest` helper in `@filbeam/retrieval`.

## What is shared

`handleFetchRequest(request, run)` registers the abort-logging listener, runs the worker handler, and converts thrown errors into responses via `handleError`. Each worker's `fetch` becomes a one-line delegation:

```js
async fetch(request, env, ctx, options) {
  return handleFetchRequest(request, () => this._fetch(request, env, ctx, options))
}
```

## Why a PR and not a direct commit

One small behavior change: `ipfs-retriever` did not previously log on request abort, and now it does, matching `piece-retriever` and `x402-piece-gateway`. It is additive logging only, no response or functional change, but worth a sign-off rather than a silent commit.

I kept each worker's thin `fetch` method rather than going further (e.g. `fetch: createFetchHandler(_fetch)`), so the diff stays on the actual change instead of re-indenting every `_fetch` body. The `_fetch` implementations are unchanged apart from `x402-piece-gateway`, whose `_fetch` now applies the `useFacilitator` default itself (the old wrapper used to supply it).

## Notes

Independent of the Tier 2/3 PRs (#674, #675), it touches the `fetch` wrapper, which those left per-worker. If more than one of these lands, the later ones rebase onto the new base.


juliangruber:
Abstraction not helpful

PR #689, opened 2026-06-25
Deduplicate ipfs-retriever candidates by service provider

### LINKED ISSUE #691: fix: deduplicate retrieval candidates by service provider in ipfs-retriever
Part of #297 

### Description

When an `ipfsRootCid` maps to multiple Filecoin pieces (shards), as is the case for Storacha content, the query `WHERE pieces.ipfs_root_cid = ?` returns one row per shard. If all shards live on the same SP (the common case), `selectRetrievalCandidate` receives **N** identical candidates and retries the same SP **N** times on failure. This gives a false impression of redundancy and wastes the retry budget without increasing resilience.

### Proposed solution

Avaliate SQL vs JS filter

### Done criteria

- [ ] `validateQueryResultsAndGetCandidates` deduplicates candidates by `serviceProviderId` after the auth cascade
- [ ] Test: 1 ipfsRootCid → 2 shards on same SP yields 1 candidate
- [ ] Test: 1 ipfsRootCid → shards on two different SPs yields 2 candidates (one per SP)
- [ ] Existing `store.test.js `tests pass unchanged

### Notes

- The sequential CAR block ordering assumption in `retrieval.js` is a separate deferred follow-up

juliangruber:
Stacked on #312.
Closes #691

Addresses https://github.com/filbeam/worker/pull/312#discussion_r3469444977.

## Problem

`ipfs-retriever` looks up candidates with `WHERE pieces.ipfs_root_cid = ?`, returning one row per piece. Under Storacha's sharding model one IPFS root CID maps to N pieces (shards) on the same service provider, so the lookup yields N candidates that differ only by piece, all with the same `serviceUrl`.

Retrieval is by IPFS root CID through the trustless gateway. `retrieveIpfsContent` uses only `serviceUrl` and `ipfsRootCid`, never the piece, so every one of those candidates is the exact same HTTP request. `selectRetrievalCandidate` then retries across all N on failure, hitting a genuinely failing provider N times before returning the 502.

## Fix

Deduplicate candidates by service provider in `validateQueryResultsAndGetCandidates`, after the authorization and quota cascade. A failing provider is now attempted once rather than once per shard, while retries across distinct providers (content replicated across providers) are preserved. This also makes the code match the existing JSDoc, which already documents "one per service provider".

The dedup lives in the `ipfs-retriever` lookup rather than the shared `selectRetrievalCandidate`, because `piece-retriever` retrieves by piece CID, where candidates sharing a `serviceUrl` are distinct requests and must not be collapsed.

### DIFF
--- a/ipfs-retriever/lib/store.js
+++ b/ipfs-retriever/lib/store.js
@@ -74,11 +74,24 @@ function validateQueryResultsAndGetCandidates(params) {
     ipfsRootCid: row.ipfs_root_cid,
   }))
 
+  // Under sharding, one IPFS root CID maps to many pieces on the same service
+  // provider, producing candidates that differ only by piece. Retrieval is by
+  // root CID and never uses the piece, so those are identical requests. Keep
+  // one candidate per service provider, so a failing provider is attempted once
+  // rather than once per shard, while still retrying across distinct providers.
+  const candidatesByServiceProvider = new Map()
+  for (const candidate of candidates) {
+    if (!candidatesByServiceProvider.has(candidate.serviceProviderId)) {
+      candidatesByServiceProvider.set(candidate.serviceProviderId, candidate)
+    }
+  }
+  const dedupedCandidates = [...candidatesByServiceProvider.values()]
+
   console.log(
-    `Validated ${candidates.length} retrieval candidate(s) for ${lookupKey} and payer '${payerAddress}'`,
+    `Validated ${dedupedCandidates.length} retrieval candidate(s) for ${lookupKey} and payer '${payerAddress}'`,
   )
 
-  return candidates
+  return dedupedCandidates
 }
 
 /**
--- a/ipfs-retriever/test/store.test.js
+++ b/ipfs-retriever/test/store.test.js
@@ -275,6 +275,87 @@ describe('getRetrievalCandidatesByWalletAndCid', () => {
       ipfsRootCid,
     })
   })
+
+  it('returns one candidate per service provider when an ipfsRootCid is sharded across many pieces', async () => {
+    const dataSetId = 'sharded-data-set'
+    const ipfsRootCid = 'sharded-ipfs-cid'
+    const payerAddress = '0x1234567890abcdef1234567890abcdef12345678'
+    const serviceProviderId = 'sharded-provider'
+
+    await withApprovedProvider(env, {
+      id: serviceProviderId,
+      serviceUrl: 'https://sharded-provider.xyz',
+    })
+
+    await env.DB.prepare(
+      'INSERT INTO data_sets (id, service_provider_id, payer_address, with_cdn, with_ipfs_indexing) VALUES (?, ?, ?, ?, ?)',
+    )
+      .bind(dataSetId, serviceProviderId, payerAddress, true, true)
+      .run()
+
+    // The same root CID is stored as three pieces (shards) on one provider.
+    for (const pieceId of ['shard-0', 'shard-1', 'shard-2']) {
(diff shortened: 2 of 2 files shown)


juliangruber:
## Can we deduplicate at the SQL level?

Technically yes, with a `GROUP BY` or a window function. I'd recommend against it here, because doing it correctly forces us to give up something the current design deliberately keeps. The reason is an ordering constraint.

### Dedup must happen after authorization, not before

A group of "same `service_provider_id`" rows for one root CID is not homogeneous. The shared query only filters on `ipfs_root_cid` (and `is_deleted`). The payer match, `with_cdn`, sanctioned check, `service_url`, and quotas are all applied later by the JS cascade. So shards on the same provider can sit in different data sets with different `payer_address`, `with_cdn`, quota, and so on.

If we `GROUP BY service_provider_id` over the raw join (picking an arbitrary representative per group) before auth, SQLite can hand us a row that fails authorization while discarding a sibling row on the same provider that would have passed, so we'd reject a valid candidate. The JS version is correct precisely because it dedups the already-authorized set: `store.js` runs the cascade, then collapses.

To dedup safely in SQL we'd have to push the whole auth and quota filter into the query:

```sql
SELECT service_provider_id, service_url, data_set_id, piece_id, ipfs_root_cid, ...
FROM pieces
JOIN data_sets ON ...
JOIN service_providers ON ...
LEFT JOIN data_set_egress_quotas ON ...
LEFT JOIN wallet_details ON ...
WHERE pieces.ipfs_root_cid = ?
  AND pieces.is_deleted IS FALSE
  AND service_providers.id IS NOT NULL AND service_providers.is_deleted IS FALSE
  AND LOWER(data_sets.payer_address) = ?
  AND data_sets.with_cdn = 1
  AND COALESCE(wallet_details.is_sanctioned, 0) = 0
  AND (NOT ? OR (COALESCE(cdn_egress_quota, 0) > 0 AND COALESCE(cache_miss_egress_quota, 0) > 0))
GROUP BY service_provider_id   -- or ROW_NUMBER() OVER (PARTITION BY service_provider_id) = 1
```

### Why that's a net loss

- We lose the cascade's granular errors. Today the JS cascade returns specific codes and messages: 404 "not indexed", 402 "no payment rail", 402 "withCDN=false", 403 "sanctioned", 402 "CDN/cache-miss quota exhausted". Once everything is a single `WHERE`, a zero-row result can't tell the client why, collapsing all of those into one generic 404 or 402.
- Bigger blast radius. `buildRetrievalCandidateQuery` is shared with `piece-retriever`, and the payer filter is currently JS-side and case-insensitive (`toLowerCase()`). Moving it into SQL changes the shared contract for both workers.
- No retrieval benefit. The problem this PR addresses is redundant HTTP retries against a failing provider. The JS dedup already eliminates that completely. SQL dedup would only shrink an already-small, already-authorized result set (the shard count for one root CID and payer), it doesn't make a single retrieval faster or more correct. The lookup cost itself is handled by the `pieces_ipfs_root_cid` index.
- SQLite-specific smell. Bare columns under `GROUP BY` pick an arbitrary row (a SQLite extension), so the portable form is a `ROW_NUMBER()` window, which is more SQL to maintain for no functional gain.

### Recommendation

Keep the dedup in JS. It's trivially correct because it runs on the post-authorization set, it preserves the specific error responses, and it fully fixes the redundant-retry problem. SQL-level dedup would only pay off if shard counts got large enough that transferring the rows mattered, and even then the right lever is the index plus maybe a `LIMIT`, not folding the auth cascade into SQL.

Issue #698, opened 2026-07-11
Indexer has received zero Goldsky webhook deliveries since 2026-06-27 ~20:01 UTC — all new pieces 404 from piece-retriever (mainnet + calibration)


beck-8:
## Summary

The Goldsky pipelines stopped delivering webhooks to the indexer workers on **both mainnet and calibration simultaneously** around **2026-06-27 20:01 UTC**. Since then, no `PieceAdded` / `DataSetCreated` / quota events have reached D1, so the piece-retriever returns 404 for every piece uploaded in the last ~2 weeks:

```
Piece_cid '<cid>' does not exist or may not have been indexed yet.
```

Originally reported by a user in filecoin-project/filecoin-pin#615 ([comment](https://github.com/filecoin-project/filecoin-pin/issues/615#issuecomment-4941420425)); I reproduced and traced it end-to-end.

## Timeline (UTC)

- **2026-06-27 20:00–20:01** — last batch of `POST /fwss/piece-added` requests received by `filcdn-indexer-mainnet` (data set 1293), all 200. All workers healthy at that moment (cron runs, Goldsky status checks, settlement all normal).
- **2026-06-27 20:01 → now** — Cloudflare telemetry shows **zero** POST requests to `index.filcdn.io` / `index.calibration.filcdn.io`. No 401s, no 4xx/5xx — complete silence on both networks. No errors at the cutoff point.
- **2026-07-11 05:11:30** — verification upload on calibration: `PieceAdded` for data set 18197 / piece 3, piece CID `bafkzcibclictjqrdjmgkhq4s4hfkl53xujhbtfstdnuy6y3ruz6zydudg3zwsgy`, tx [`0x47830d…8cbe`](https://calibration.filfox.info/en/message/0x47830d6c23db2dbcc394f8f88baaefda231b67a9b7188a72a9626c044e968cbe), block 3880677, emitted by FWSS `0x0292…7CA0` (the address the subgraph watches).
- **+1 min** — event queryable in the public `filbeam-calibnet/1.1.0` subgraph GraphQL endpoint (head lag ~49s, `hasIndexingErrors: false`).
- **+21 min** — piece-retriever still 404 (polled every 30s); no webhook ever arrived at the indexer.

## Ruled out

- **Worker code/config** — no relevant changes around June 27; `subgraph/pipelines/*.yaml` untouched since January; indexer workers demonstrably alive.
[...]

PR #699, opened 2026-07-11
feat(indexer): add /health endpoint verifying webhook delivery from Goldsky pipelines

beck-8:
Adds an unauthenticated `GET /health` to the indexer that verifies events indexed by the Goldsky subgraph are actually being delivered to this worker and written to D1 — the gap that let #698 go unnoticed for two weeks (subgraph lag monitoring stayed green while pipeline→webhook delivery was down).

**How it works:** fetch the newest `PieceAdded` from the subgraph that is older than `DELIVERY_LAG_BUFFER_SECONDS` (new var, default 1800), and require a matching row in the `pieces` table. Row presence proves delivery even for later-removed pieces, since removals upsert `is_deleted=TRUE` — so no schema change is needed.

| Status | HTTP | Meaning |
|---|---|---|
| `healthy` | 200 | newest eligible piece is in D1 |
| `idle` | 200 | no eligible on-chain events to verify |
| `delivery_lagging` | 503 | newest eligible piece missing from D1 |
| `d1_unhealthy` | 503 | D1 query failed |
| `subgraph_unknown` | 503 | subgraph unreachable / GraphQL errors / unusable response |

Only exact `GET /health` bypasses the webhook secret; all other routes keep existing auth. Responses expose only public on-chain identifiers; failures are logged via `console.warn`.

Verified end-to-end with `wrangler dev` against the live calibnet subgraph: empty D1 → `delivery_lagging` [503]; delivering the flagged piece through the authenticated `POST /fwss/piece-added` route → `healthy` [200]. (The 503 currently reflects reality — see #698.)

**Known limits (by design, MVP):** a single-point probe detects full delivery stalls, not gaps in older backfill; no verification is possible during zero-activity windows (`idle`). Follow-up: a Gatus check on `/health` (`[STATUS] == 200`) will be added in the infra repo.


indexer/lib/health.js
```diff
@@ -0,0 +1,143 @@
...
+  if (pieceAddeds.length === 0) {
+    return { status: 'idle' }
+  }
+
+  const [piece] = pieceAddeds
+  /** @type {unknown} */
+  let row
+  try {
+    row = await env.DB.prepare(
+      'SELECT 1 FROM pieces WHERE id = ? AND data_set_id = ?',
```
rjan90: Could we require `cid IS NOT NULL` here? `removeDataSetPieces` can create a tombstone containing only `id`, `data_set_id`, and `is_deleted`, so the current query reports `healthy` even if the `PieceAdded` webhook never arrived and only the removal did. `insertDataSetPiece` always populates `cid`, including after an out-of-order removal, so checking it would distinguish these cases. The “delivered and later removed” test should also seed a CID, with a separate test confirming that a removal-only tombstone remains `delivery_lagging`.

rjan90 (approved):
Looks good overall. I left one non-blocking inline follow-up for the removal-only tombstone case.

Issue #694, opened 2026-07-01
[Complete by 2026-07-07] Restore filcdn.com nameserver delegation to Cloudflare


BigLep:
## Problem

Cloudflare reported on July 1, 2026 that `filcdn.com` is no longer using Cloudflare nameservers and therefore no longer benefits from Cloudflare's network.

The domain is currently delegated to:

- `ns3.expirationwarning.net`
- `ns7.expirationwarning.net`

Cloudflare expects it to be delegated to:

- `aspen.ns.cloudflare.com`
- `jacob.ns.cloudflare.com`

The Cloudflare zone is now in a **moved** state and, unless it has an active paid subscription, Cloudflare says it will be automatically deleted after 7 days.

## Action needed

- Identify the registrar/DNS provider for `filcdn.com`.
- Update the domain's authoritative nameservers to the two Cloudflare nameservers above.
- Confirm the delegation has propagated.
- Verify that the zone returns to active status in the [Cloudflare dashboard](https://dash.cloudflare.com/37573110e38849a343d93b727953188f/filcdn.com).
- Verify `filcdn.com` and relevant production hostnames resolve and serve traffic correctly.

## Source

<img width="726" height="1184" alt="Image" src="https://github.com/user-attachments/assets/1114132d-c1aa-40cd-8979-92969182bb59" />

BigLep:
@rjan90 or @TippyFlitsUK: could you please take this?

TippyFlitsUK:
I don't believe we ever transferred these, @BigLep. @rjan90, did we have transfer codes for this domain as well?

rvagg:
<img width="1000" height="374" alt="Image" src="https://github.com/user-attachments/assets/3cb804e2-d06d-4718-8d76-eb8e64e241c7" />

gotta get that domain and renew it cause we're about to lose it .. it's currently non-functioning so nobody's filcdn requests will be working.

rjan90:
Reached out to Julian / Miro to understand where this domain is, and if this is legacy/not needed anymore. There is at least a runtime redirect https://github.com/filbeam/worker/blob/ec6eff42f82a67ab3bc482676efb471551cc7654/piece-retriever/bin/piece-retriever.js#L54-L57, redirecting *.[filcdn.io](http://filcdn.io/) requests to *.[filbeam.io](http://filbeam.io/)

rjan90:
Update after chatting with Miro about this:
[...]

PR #700, opened 2026-07-11
fix: migrate Goldsky webhook sinks to filbeam.io

rjan90:
## Summary

- add exact `filbeam.io` Worker routes for the calibration and mainnet indexers
- migrate all 11 Goldsky webhook sinks per environment from the retired `filcdn.io` hosts to `filbeam.io`
- preserve the existing pipeline sources, sink keys, authentication secret, and delivery settings

## Root cause

The legacy `filcdn.io` domain was intentionally retired and removed from Cloudflare, but the Goldsky pipeline definitions still used `index.calibration.filcdn.io` and `index.filcdn.io`. When DNS delegation disappeared, both pipelines continued reading source records but could no longer connect to their webhook sinks, causing the delivery outage described in #698.

## Rollout

1. Deploy the calibration indexer route and verify an unauthenticated POST to `index.calibration.filbeam.io` returns `401` from the indexer rather than `405` from the piece-retriever wildcard.
2. Apply the updated Calibnet pipeline from its last snapshot and verify backlog recovery, including data set `18197`, piece `3`.
3. Only after Calibnet is healthy, repeat the route deployment and pipeline update for mainnet.

Refs #698 and #694.

## Validation

- `npm run lint:fix`
- `npm test` — 34 test files, 381 tests passed
- `npx wrangler deploy --dry-run --env calibration --config indexer/wrangler.toml`
- `npx wrangler deploy --dry-run --env mainnet --config indexer/wrangler.toml`

### DIFF
--- a/indexer/wrangler.toml
+++ b/indexer/wrangler.toml
@@ -74,6 +74,10 @@ binding = "RETRY_QUEUE"
 binding = "PROCESSED_EVENTS"
 id = "b18c3c10-fake-dev-id"
 
+[env.calibration.route]
+pattern = "index.calibration.filbeam.io/*"
+zone_name = "filbeam.io"
+
 [env.calibration.vars]
 ENVIRONMENT = "calibration"
 WALLET_SCREENING_BATCH_SIZE = 1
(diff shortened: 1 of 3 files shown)


beck-8:
There seems to be no problem.

BravoNatalie (approved):

